PR #15, opened 2023-04-19
feat: cabi-realloc-adapter support

guybedford:
WIP fix to use the new `cabi_realloc_adapter` export binding to separate adapter allocations from component allocations allowing for easier management of a stack-like allocator (which we currently use for convenience).

Pending https://github.com/bytecodealliance/wasm-tools/pull/989.


guybedford:
This has been rolled into the toolchain upgrade in https://github.com/bytecodealliance/componentize-js/pull/18.

PR #17, opened 2023-04-19
add example for imports and exports

Gentle:
I added an updated example that works with the current wasmtime version

also added an example host function that gets implemented in rust and called by the js component (I adapted this example from the livestream you did a few months ago)


guybedford:
Thank you for the PR, you're completely right we should maintain an end-to-end example here that is regularly tested for this. Just updated the workflow today with https://github.com/bytecodealliance/componentize-js/pull/21 as well.

If you'd like to update the example, I'll merge it in, otherwise I will follow up with the update here this week.

guybedford:
Landed in https://github.com/bytecodealliance/componentize-js/pull/23.

PR #21, opened 2023-05-08
docs: update example

guybedford:
The example workflow has changed, this updates the Wasmtime configurations to ones that will work with componentize.

Unfortunately the WASI bindings require an older version of preview2, so there is currently a WASI bindings failure on this pending to fix before we merge.

We either need to trawl through the release matrix to work this out, otherwise I can aim to get the latest preview2 update to componentize-js soon which should resolve this.

### DIFF
--- a/EXAMPLE.md
+++ b/EXAMPLE.md
@@ -22,6 +22,7 @@ export function hello (name) {
 
 The component can then be built with the `componentize` API:
 
+componentize.mjs
 ```js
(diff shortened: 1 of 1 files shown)


squillace (commented):
you didn't ask for my approval, but this I confirm does compile but do not run yet. I understand you're working on an update to preview2, so I think with that update things should work.

Issue #68, opened 2023-10-30
Resource destructors via Symbol.dispose


guybedford:
We should support resource destructors for the ComponentizeJS resources implementation.

Specifically - `Symbol.dispose` allows resources authored in JS to define a destructor. And `Symbol.dispose` also provides a mechanism for **early destruction** of **own** resources, instead of having to wait for the GC.

In addition, the finalization registry implemented for resources doesn't currently cater to `dtor` calls, which can be looked up with a custom parse of the component Wasm.

guybedford:
As far as I can tell, the `dtor` information is not actually available to the wit parser types though, and requires a custom parse of the component like in  https://github.com/bytecodealliance/wasmtime/blob/5062e34/crates/environ/src/component/translate.rs#L421.

Unless there is a way to get this from the bindgen types that I don't know about? (cc @alexchricton)

PR #69, opened 2023-11-02
support destructors for imported resources

dicej:
This provides partial support for #68.  I've focused on imported resources to begin with since APIs such as `wasi:http` make heavy use of child handles, which must be disposed in the correct order to avoid traps.

Supporting destructors for exported resources may also be useful, but not as urgent with respect to WASI.

### DIFF
--- a/crates/spidermonkey-embedding-splicer/src/bindgen.rs
+++ b/crates/spidermonkey-embedding-splicer/src/bindgen.rs
@@ -372,6 +372,11 @@ pub fn componentize_bindgen(resolve: &Resolve, id: WorldId, name: &str) -> Compo
 }
 
 impl JsBindgen<'_> {
+    fn intrinsic(&mut self, intrinsic: Intrinsic) -> String {
+        self.all_intrinsics.insert(intrinsic);
+        return intrinsic.name().to_string();
+    }
+
     fn exports_bindgen(&mut self) {
         for (key, export) in &self.resolve.worlds[self.world].exports {
             let name = self.resolve.name_world_key(key);
@@ -516,8 +521,32 @@ impl JsBindgen<'_> {
                             );
                         }
 
-                        if resource.is_some() {
-                            uwriteln!(self.src, "}}");
+                        if let Some(ty) = resource {
+                            let lower_camel = &self.resolve.types[ty]
+                                .name
+                                .as_ref()
+                                .unwrap()
+                                .to_lower_camel_case();
+
+                            let prefix = iface_name
+                                .as_deref()
+                                .map(|s| format!("{s}$"))
+                                .unwrap_or(String::new());
+
+                            let resource_symbol = self.intrinsic(Intrinsic::SymbolResourceHandle);
+                            let dispose_symbol = self.intrinsic(Intrinsic::SymbolDispose);
+
+                            uwriteln!(
+                                self.src,
+                                "
+                                    [{dispose_symbol}]() {{
+                                        finalizationRegistry_import${prefix}{lower_camel}.unregister(this);
+                                        $resource_import${prefix}drop${lower_camel}(this[{resource_symbol}]);
+                                        this[{resource_symbol}] = null;
+                                    }}
+                               }}
+                               "
+                            );
                         }
                     }
                 }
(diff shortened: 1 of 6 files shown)


dicej:
Here's an example: https://github.com/fermyon/spin-fileserver/blob/main/examples/javascript/app.mjs

crates/spidermonkey-embedding-splicer/src/bindgen.rs
```diff
@@ -516,8 +521,32 @@ impl JsBindgen<'_> {
...
+                            let resource_symbol = self.intrinsic(Intrinsic::SymbolResourceHandle);
+                            let dispose_symbol = self.intrinsic(Intrinsic::SymbolDispose);
+
+                            uwriteln!(
+                                self.src,
+                                "
+                                    [{dispose_symbol}]() {{
+                                        finalizationRegistry_import${prefix}{lower_camel}.unregister(this);
+                                        $resource_import${prefix}drop${lower_camel}(this[{resource_symbol}]);
+                                        delete this[{resource_symbol}];
```
guybedford, line 545: To nitpick, shape changes in JS have a very slight performance cost as the object gets copied into the new shape. Setting a property to `null` is therefore faster from a microbenchmark perspective, and that would also avoiding the need for the configurability PR.

crates/spidermonkey-embedding-splicer/src/bindgen.rs
```diff
@@ -516,8 +521,32 @@ impl JsBindgen<'_> {
...
+
+                            let resource_symbol = self.intrinsic(Intrinsic::SymbolResourceHandle);
+                            let dispose_symbol = self.intrinsic(Intrinsic::SymbolDispose);
+
+                            uwriteln!(
+                                self.src,
+                                "
+                                    [{dispose_symbol}]() {{
+                                        finalizationRegistry_import${prefix}{lower_camel}.unregister(this);
+                                        $resource_import${prefix}drop${lower_camel}(this[{resource_symbol}]);
```
guybedford, line 544: Do we need to separate the borrow case here? Or should we be dropping borrows as well? The finalization registry avoids a double drop, but are we not already explicitly dropping borrows at the call end?

guybedford (commented):
Amazing, thanks for getting to this. A couple of comments.

crates/spidermonkey-embedding-splicer/src/bindgen.rs
dicej: Sounds good.  I'll push an update.

crates/spidermonkey-embedding-splicer/src/bindgen.rs
guybedford: I don't currently have a very clear picture of the resource model, but some of my questions are:
* Currently, without this PR, are all borrows dropped via the finalization registry? Does that mean that they are not dropped at the end of the call as well? And if there is a drop of a borrow at the end of the call, if a borrow is dropped by the finalization registry before that end of the call, does that create a double drop or are there double drop protections in this case?
* Then secondly, with this PR, would the same not apply, where calling symbol.dispose might pre-empt the default end-of-call drop that happens?

Also I may well be misinterpreting the model, resources are complex, so let me know if there are any gaps in my interpretation too.

I don't think we should have separate classes, but just that I would have thought that the finalization registry would only be needed for owns and not borrows in general (which is what jco does on the host side).

Perhaps this is more an implementation item separate to this PR though?

dicej: Well, now I'm confused.  I just pushed a commit that expands the `resource-borrow-import` test to cover a couple of relevant cases, but I can't figure out how to make them fail.  The finalizer function doesn't seem to be run, and yet the host doesn't complain the the borrow handle is not dropped.  Perhaps Jco isn't as picky as it should be about whether the guest drops its borrow handles before returning?

guybedford: Ok, it sounds like this is an existing issue then - let's track this further in https://github.com/bytecodealliance/componentize-js/issues/70 and https://github.com/bytecodealliance/jco/issues/249, and get this PR landed in the mean time.

PR #86, opened 2024-02-22
fix(example): Update to wasmtime 17

kyleconroy:
I ran into a bunch of issues trying to use this example with a component from [componentize-py](https://github.com/bytecodealliance/componentize-py). It turns out I just needed to upgrade to wasmtime 17.


guybedford:
Thanks for the PR!

Issue #104, opened 2024-05-06
Question - Should componentize-js tree shake unused imports? 


karthik2804:
Should the componentizing process perform tree shaking to exclude the wasm imports not imported by the JS guest code? 

An example use case is when a npm package is written for an SDK, the wit files included with it will have all the possible imports. The guest code might choose not to use all the available imports, it would be good to only generate the bindings for the imports that are actually used. It will help reduce the binary size and also could help with running the component on hosts that might not support all the imports defined in the SDK.

guybedford:
This sounds like a great idea!

karthik2804:
As a follow-up to the [comment](https://github.com/bytecodealliance/ComponentizeJS/pull/109#issuecomment-2138402093), imagine the target world specifies all imports an SDK package might want to use. When a user decides to use a particular subset of the target world imports, we could potentially match the imports in the guest code using the [following code](https://github.com/bytecodealliance/ComponentizeJS/blob/main/src/componentize.js#L124) to the ones in the target world. Anything in the target world not used by the guest could be stubbed out as `unreachable`.  This would limit the number of imports that a host would have to satisfy to run the component. 

I am imagining another function in addition to `splice_bindings` and `stub_wasi` that we could optionally call to stub imports in the target world that are not used. 

Let me know if this makes sense. I could try to create a PoC after #109.

guybedford:
Ah nice, yeah that sounds ideal I agree!

karthik2804:
The first thing I ran into was, It is not straightforward to track the implicit imports that get brought in. For example importing `wasi:cli/stdin@0.2.0` in the JS guest code would also mean we need to import `wasi:io`. This is something I have to think about a bit more.

guybedford:
I think we only need to do this for the direct imports in the splicer itself, where we synthesize all the core imports, where we would want to skip unused imports. This can be done by providing the filtered list of imports to https://github.com/bytecodealliance/ComponentizeJS/blob/main/crates/spidermonkey-embedding-splicer/src/splice.rs#L38, perhaps with a new argument "used_imports" to the main [splice bindings](https://github.com/bytecodealliance/ComponentizeJS/blob/main/crates/spidermonkey-embedding-splicer/src/lib.rs#L107) function.

The component model already has the "gc" code to prune unused transitive imports, so the rest should follow simply from there.

karthik2804:
completed with #117

PR #109, opened 2024-05-27
do not stub imports that are part of the target world

### LINKED ISSUE #105: Question - Can we optionally enable `cli` and `fs` worlds.
Could we optionally enable the `cli` and `fs` worlds. This would enable building compatibility layers such that more libraries would work and also improve the variety of applications that can be built. 

Are there any concerns with this? if not, I am happy to create a PR.

karthik2804:
This PR makes sure that if an import is present in the target world, it does not get stubbed out. I went with a different approach than what was suggested in the [issue comment](https://github.com/bytecodealliance/ComponentizeJS/issues/105#issuecomment-2110865818) as I noticed the module only contained the wasi imports if they are a part of the target world and was simpler than having to prepare the `wit`. Let me know if I am missing something. 

fixes #105

### DIFF
--- a/crates/spidermonkey-embedding-splicer/src/stub_wasi.rs
+++ b/crates/spidermonkey-embedding-splicer/src/stub_wasi.rs
@@ -1,12 +1,17 @@
 use anyhow::{bail, Result};
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::{
+    collections::HashSet,
+    path::PathBuf,
+    time::{SystemTime, UNIX_EPOCH},
+};
 use walrus::{
     ir::{BinaryOp, MemArg, StoreKind, UnaryOp},
     Function, FunctionBuilder, FunctionId, FunctionKind, ImportKind, ImportedFunction, InitExpr,
     InstrSeqBuilder, LocalId, Module, ValType,
 };
+use wit_parser::{Resolve, UnresolvedPackage};
 
-use crate::Features;
+use crate::{parse_wit, Features};
 
 fn stub_import<StubFn>(
     module: &mut Module,
@@ -52,12 +57,40 @@ fn unreachable_stub(body: &mut InstrSeqBuilder) -> Result<Vec<LocalId>> {
     Ok(vec![])
 }
 
-pub fn stub_wasi(wasm: Vec<u8>, features: Vec<Features>) -> Result<Vec<u8>> {
+pub fn stub_wasi(
+    wasm: Vec<u8>,
+    features: Vec<Features>,
+    wit_source: Option<String>,
+    wit_path: Option<String>,
+    world_name: Option<String>,
+) -> Result<Vec<u8>> {
+    let (resolve, id) = if let Some(wit_source) = wit_source {
+        let mut resolve = Resolve::default();
+        let path = PathBuf::from("component.wit");
+        let pkg = UnresolvedPackage::parse(&path, &wit_source)?;
+
+        let id = resolve.push(pkg)?;
+
+        (resolve, id)
+    } else {
+        parse_wit(&PathBuf::from(wit_path.unwrap()))?
+    };
+
+    let world = resolve.select_world(id, world_name.as_deref())?;
+
+    let target_world = &resolve.worlds[world];
+    let mut target_world_imports = HashSet::new();
+
+    for (key, _) in &target_world.imports {
+        target_world_imports.insert(resolve.name_world_key(key));
+    }
+
     let mut module = Module::from_buffer(wasm.as_slice())?;
 
     stub_preview1(&mut module)?;
-    stub_filesystem(&mut module)?;
-    stub_cli(&mut module)?;
+
+    stub_filesystem(&mut module, &target_world_imports)?;
+    stub_cli(&mut module, &target_world_imports)?;
 
     if !features.contains(&Features::Random) {
         stub_random(&mut module)?;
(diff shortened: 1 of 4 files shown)


guybedford:
There are two ways that imports can be included in the output binary:

1. Imports are pulled in by the engine itself
2. World imports being converted into imports by the splicing code

The stubbing code here acts on the binary that is the union of the above. Therefore the stubbing code cannot know for a given import if it was used by the engine or if it is used by the target world.

Therefore saying that features only stub out when the final binary doesn't use them excludes engine imports pulled in that we _do_ want to stub. For example libc builtins that get pulled in that are unused.

This means that this PR risks outputting a binary with extra engine imports that were supposed to be stubbed. I think you will find that the "pure component" case of no features is thus no longer working with this PR, and that pure components have imports again.

This list of these engine bulitins that should always be stubbed can be determined by building a binary against any non-WASI world. The problem is that this list should be excluded against any WASI imports of the world being targetted. And thus we need to parse the target world to do the exclusion as I described in https://github.com/bytecodealliance/ComponentizeJS/issues/105#issuecomment-2110865818.

I'm open to counter arguments if you can prove any of the above otherwise.

crates/spidermonkey-embedding-splicer/src/stub_wasi.rs
```diff
@@ -82,7 +100,9 @@ pub fn stub_wasi(wasm: Vec<u8>, features: Vec<Features>) -> Result<Vec<u8>> {
         stub_io(&mut module)?;
```
guybedford, line 100: We probably need to extend `has_io` to also include `|| module_imports.contains("wasi:filesystem") || module_imports.contains("wasi:sockets")`.

guybedford:
I just checked this PR and can confirm the pure component case is still okay, but the edge cases are still not quite consistent I think. To dig into the nuances:

* Preview1 builtins are always stubbed (`stub_preview1()`)
* It just so happens that filesystem, CLI and sockets are not at all used by StarlingMonkey builtins, so as a result we will never have this ambiguous union case with world imports. If StarlingMonkey builtins ever bind to these subsystems in future we will get these ambiguity cases.
* This then leaves random, CLI, clocks and HTTP as cases where having the feature disabled will still stub out the world. This results in the case where you must set `features: ['random']` if you are binding to a world that uses random.

So the consequence is that for example having HTTP disabled by default will result in not being able to output to an HTTP world, unless you explicitly enable the HTTP feature as well.

Right now, features is something that relates to the engine itself, and gets unioned with the target world. It seems like it would be odd to have to align the features usage with the exact target world usage.

For example consider if we add support for websockets in future. This is a "websockets" feature. But whether websockets is enabled or not shouldn't affect weather I can target a world that uses websockets.

karthik2804:
I think the above tracks in my head. I am happy to take a stab at using an approach similar to splicing if that is the recommended approach.

guybedford:
Thanks for being open to that! It's just a few lines of code. An example of walking the exports of a WIT world is here - https://github.com/bytecodealliance/ComponentizeJS/blob/main/crates/spidermonkey-embedding-splicer/src/lib.rs#L143. Where instead of `exports`, `imports` can be used. Those strings can then be directly used as the exclude strings when performing stubbing.

karthik2804:
I have updated the code to resolve the target world imports and use that to conditionally stub imports. I wanted to check if this is the desired approach. If this is on the right approach I will rebase the PR.

crates/spidermonkey-embedding-splicer/src/stub_wasi.rs
```diff
@@ -77,16 +117,36 @@ pub fn stub_wasi(wasm: Vec<u8>, features: Vec<Features>) -> Result<Vec<u8>> {
...
+        || target_world_imports.contains("wasi:sockets/network@0.2.0")
+        || target_world_imports.contains("wasi:filesystem/types@0.2.0")
+        || target_world_imports.contains("wasi:filesystem/preopens@0.2.0")
+        || target_world_imports.contains("wasi:cli/environment@0.2.0")
+        || target_world_imports.contains("wasi:cli/exit@0.2.0")
+        || target_world_imports.contains("wasi:cli/terminal-stdin@0.2.0")
+        || target_world_imports.contains("wasi:cli/terminal-stdout@0.2.0")
+        || target_world_imports.contains("wasi:cli/terminal-stderr@0.2.0")
+        || target_world_imports.contains("wasi:cli/terminal-input@0.2.0")
+        || target_world_imports.contains("wasi:cli/terminal-output@0.2.0");
```
guybedford, line 147: `cli/environment` and `cli/exit` actually don't require IO:

```suggestion
    return target_world_imports.contains("wasi:sockets/instance-network@0.2.0")
        || target_world_imports.contains("wasi:sockets/udp@0.2.0")
        || target_world_imports.contains("wasi:sockets/udp-create-socket@0.2.0")
        || target_world_imports.contains("wasi:sockets/tcp@0.2.0")
        || target_world_imports.contains("wasi:sockets/tcp-create-socket@0.2.0")
        || target_world_imports.contains("wasi:sockets/ip-name-lookup@0.2.0")
        || target_world_imports.contains("wasi:sockets/network@0.2.0")
        || target_world_imports.contains("wasi:filesystem/types@0.2.0")
        || target_world_imports.contains("wasi:filesystem/preopens@0.2.0")
        || target_world_imports.contains("wasi:cli/terminal-stdin@0.2.0")
        || target_world_imports.contains("wasi:cli/terminal-stdout@0.2.0")
        || target_world_imports.contains("wasi:cli/terminal-stderr@0.2.0")
        || target_world_imports.contains("wasi:cli/terminal-input@0.2.0")
        || target_world_imports.contains("wasi:cli/terminal-output@0.2.0");
```

guybedford (approved):
Amazing, thank you for putting this together, LGTM.

guybedford:
Thinking about this some more: I suppose an alternative here would just be to stub the WASI imports **before** doing the splicing, ie before the engine gets passed in here - https://github.com/bytecodealliance/ComponentizeJS/blob/main/src/componentize.js#L52.

Then any world imports would be fully additive without needing to go through the world exclusion process.

This also seems like it would work better with composition workflows where right now as soon as something is used it is no longer stubbed.

Sorry for being slow on this one, took me a minute to think it through.

karthik2804:
No worries. I will take a stab at updating it to the above approach tomorrow.

On another note,  Working on this also made me realize we can potentially solve #104 similarly. We could use `@babel/parser` and `@babel/traverse` to parse the source file and stub any of the imports not in use by the guest code but part of the target world.

guybedford:
Definitely share in more detail what you're thinking about the stubbing in that issue. Note that we already use [this lexer](https://github.com/guybedford/es-module-lexer) to analyze imports and exports in https://github.com/bytecodealliance/ComponentizeJS/blob/main/src/componentize.js#L124.

karthik2804:
Is the suggested idea of calling `stubWasi` before `spliceBindings` something like the following? 

```js
  const features = []; // Intionally  leaving out the part of checking for enabled features.
  const stubbedEngine = stubWasi(await readFile(engine), features);

  let { wasm, jsBindings, importWrappers, exports, imports } = spliceBindings(
    sourceName,
    stubbedEngine,
    witWorld,
    maybeWindowsPath(witPath),
    worldName,
    false
  );
```

When using that to build the example I get the following, 
```bash
Error: calling the Reactor initialization function

Caused by:
    0: error while executing at wasm backtrace:
           0: 0x7fbf - <unknown>!<wasm function 20>
           1: 0x76b206 - <unknown>!<wasm function 11512>
    1: wasm trap: wasm `unreachable` instruction executed
file:///Users/karthik_ganeshram/Work/wasm-tools-contrib/ComponentizeJS/src/componentize.js:219
    throw new Error(err);
          ^

Error: Failed to initialize the compiled Wasm binary with Wizer:
Wizering failed to complete
    at componentize (file:///Users/karthik_ganeshram/Work/wasm-tools-contrib/ComponentizeJS/src/componentize.js:219:11)
    at async file:///Users/karthik_ganeshram/Work/wasm-tools-contrib/ComponentizeJS/example/componentize.js:7:23

Node.js v21.7.3
```
I will look into it more.

guybedford:
Right, stubbing can't happen before Wizering. And there is some trickiness around environment variables and FS access here, where those subsystems would need to be specifically stubbed as Wizer stubs. Difficult to come up with a consistent model here! Perhaps the approach you implemented here is in fact a good one for now.

I can review and land if you like in the mean time, before we explore alternatives further.

karthik2804:
It would be great to review and land this, it would help with the usage of extension libraries in JS to provide node APIs until we get them directly from `starlingmonkey`. I can fix the above suggestion and also remove the `unwraps` in a bit.

karthik2804:
This should now be ready for a review.

PR #111, opened 2024-05-28
deps: update to latest starlingmonkey

guybedford:
Updates to the latest version of StarlingMonkey as of https://github.com/bytecodealliance/StarlingMonkey/commit/d7c63482392bcde2882503f77d4ad75f85f36d50.


karthik2804:
When building this PR and running the tests using `npm run build && npm run test` I get the following errors on every binding test.

```
Splicer error: unable to read prelude sequence, continuing for debug build but note binding functions will not work!
```

guybedford:
@karthik2804 make sure you are definitely not using a debug build or referencing the debug build in this project, as that error is only associated with debug builds.

karthik2804:
I just attempted it with a fresh clone of the project with the following commands and I get the same error

```bash
git clone https://github.com/bytecodealliance/ComponentizeJS
cd ComponentizeJS
gh pr checkout 111
npm install
git submodule update --init --recursive
npm run build && npm run test
```

guybedford:
Ahh thanks for spotting that, the splicing functions might need to be updated to the latest binary format here then due to the Spidermonkey upgrade.

karthik2804:
I did a little print line debugging on https://github.com/bytecodealliance/ComponentizeJS/blob/519256f5632f48e42cfea54bfd588f383eccb855/crates/spidermonkey-embedding-splicer/src/splice.rs#L231

The snippet below is a part of the output. 

```rust
InstrSeq {
    ...
    instrs: [
        (
            LocalGet(
                LocalGet {
                    local: Id {
                        idx: 21481,
                    },
                },
            ),
            ...
}
```

guybedford:
@karthik2804 I fixed up the splicing portion of this PR so those tests should be passing now, if that helps.

karthik2804:
@guybedford Thank you! I can confirm that the tests are passing now. I will now attempt to see if I can get fetch to work when building an app with componentize-js.

karthik2804:
@guybedford Looking at the diff with respect to starlingmonkey it no longer includes the fetch rework unless I am missing something. 
https://github.com/bytecodealliance/StarlingMonkey/compare/c1d7439a3412496db0c5d4004d723396cc46a892...5bdb946f137e946206fb865cdba92b8fb0654416

guybedford:
@karthik2804 it needs a merge of that work against https://github.com/bytecodealliance/StarlingMonkey/pull/56 now, I've tried to add that, let's see how it looks in CI.

karthik2804:
Fetch seems to be supported but it does look like `async` is broken. The guest appears to be terminating on the first `await` call. A minimal reproduction can be achieved by modifying the sample in the `examples/` folder of this repo. 

```js
// hello.js
export function hello(name) {
  console.log("in here")
  test()
  return `Hello ${name}`;
}

async function test() {
  console.log("in test")
  await new Promise(resolve => setTimeout(resolve, 0));
  console.log("after await")
}
```

guybedford:
@karthik2804 I've posted https://github.com/bytecodealliance/ComponentizeJS/issues/112 with an outline of an approach, happy to discuss further if it helps.

guybedford:
@karthik2804 here's what I do to implement async in the mean time - a `run` and `ready` polling implemention is used in the tests here - https://github.com/bytecodealliance/ComponentizeJS/blob/main/test/test.js#L25.

karthik2804:
@guybedford Thank you for the example implementation. I am trying to target the  `incomingHandler` export of `wasi:http` where I do not believe this polling implementation applies since I am not in control of the host side implementation/`wit`.

guybedford:
In that case, an implementation along the lines of #112 sounds like the right kind of approach to me.

PR #116, opened 2024-06-17
add option to retain fetchEvent handler

### LINKED ISSUE #114: Optionally allow using `fetchEvent` to handle `wasi:http/incomingHandler@0.2.0#handle`
As the title suggests, it would optionally use the `fetchEvent` that Starlingmonkey provides to handle incoming `wasi-http` requests. The upside to this would be to write apps that target a more idiomatic API which developers will be familiar with.  

https://github.com/bytecodealliance/ComponentizeJS/blob/63e3453bab9513296ad3ec65607540288487bf93/crates/spidermonkey-embedding-splicer/src/splice.rs#L56-L63

karthik2804:
This PR adds the ability to use the `fetchEvent` Handler provided by `StarlingMonkey`. 

fixes #114

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,10 @@ Setting `disableFeatures: ['random', 'stdio', 'clocks']` will disable all featur
 
 Note that pure components **will not report errors and will instead trap**, so that this should only be enabled after very careful testing.
 
+The features that are not included by default are:
+* `'http'` - Support for sending and receiving HTTP requests, depends on `wasi:io` 
+* `'fetch-event'` - Enables using `fetchEvent` to respond to `wasi:http/incoming-handler@0.2.0#handle`. If the target world does note export `wasi:http/incoming-handler@0.2.0`, this will be ignored.
+
 Note that features explicitly imported by the target world cannot be disabled - if you target a component to a world
 that imports `wasi:clocks`, then `disableFeatures: ['clocks']` will not be supported.
 
--- a/src/componentize.js
+++ b/src/componentize.js
@@ -45,6 +45,7 @@ export async function componentize(jsSource, witWorld, opts) {
     worldName,
     disableFeatures = [],
     enableFeatures = [],
+    retainFetchEvent = false
   } = opts || {};
 
   let { wasm, jsBindings, importWrappers, exports, imports } = spliceBindings(
@@ -53,7 +54,8 @@ export async function componentize(jsSource, witWorld, opts) {
     witWorld,
     maybeWindowsPath(witPath),
     worldName,
-    false
+    false,
+    retainFetchEvent
   );
 
   // we never disable a feature that is already in the target world usage
--- a/src/componentize.js
+++ b/src/componentize.js
@@ -45,8 +45,9 @@ export async function componentize(jsSource, witWorld, opts) {
     worldName,
     disableFeatures = [],
     enableFeatures = [],
-    retainFetchEvent = false
   } = opts || {};
+
+  const retainFetchEvent = enableFeatures.includes('fetch-event')
 
   let { wasm, jsBindings, importWrappers, exports, imports } = spliceBindings(
     sourceName,
--- a/crates/spidermonkey-embedding-splicer/src/bindgen.rs
+++ b/crates/spidermonkey-embedding-splicer/src/bindgen.rs
@@ -124,7 +124,12 @@ pub struct Componentization {
     pub resource_imports: Vec<(String, String, u32)>,
 }
 
-pub fn componentize_bindgen(resolve: &Resolve, id: WorldId, name: &str) -> Componentization {
+pub fn componentize_bindgen(
+    resolve: &Resolve,
+    id: WorldId,
+    name: &str,
+    retain_fetch_event: bool,
+) -> Componentization {
     let mut bindgen = JsBindgen {
         src: Source::default(),
         esm_bindgen: EsmBindgen::default(),
@@ -342,6 +347,7 @@ pub fn componentize_bindgen(resolve: &Resolve, id: WorldId, name: &str) -> Compo
         "$source_mod",
         &mut bindgen.local_names,
         name,
+        retain_fetch_event,
     );
 
     let js_intrinsics = render_intrinsics(&mut bindgen.all_intrinsics, false, true);
@@ -1001,6 +1007,7 @@ impl EsmBindgen {
         imports_object: &str,
         _local_names: &mut LocalNames,
         source_name: &str,
+        retain_fetch_event: bool,
     ) {
         // TODO: bring back these validations of imports
         // including using the flattened bindings
@@ -1043,6 +1050,10 @@ impl EsmBindgen {
                 ");
         }
         for (export_name, binding) in &self.exports {
+            // If we are retaining the fetch event, then we should just let StarlingMonkey handle the event directly.
+            if retain_fetch_event && export_name == "wasi:http/incoming-handler@0.2.0" {
+                continue;
+            }
             match binding {
                 Binding::Interface(bindings) => {
                     uwrite!(output, "const ");
--- a/crates/spidermonkey-embedding-splicer/src/lib.rs
+++ b/crates/spidermonkey-embedding-splicer/src/lib.rs
@@ -1,6 +1,9 @@
 use anyhow::{bail, Context, Result};
 use bindgen::BindingItem;
-use std::path::{Path, PathBuf};
+use std::{
+    collections::HashSet,
+    path::{Path, PathBuf},
+};
 
 mod bindgen;
 mod splice;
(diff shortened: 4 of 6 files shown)


src/componentize.js
```diff
@@ -45,6 +45,7 @@ export async function componentize(jsSource, witWorld, opts) {
     worldName,
     disableFeatures = [],
     enableFeatures = [],
+    retainFetchEvent = false
```
guybedford, line 48: I wonder if it would make sense to integrate this into the `enableFeatures` model and call it `fetch-event`?

guybedford (commented):
Can we also add this to the readme?

src/componentize.js
karthik2804: Oooh! I like the idea of adding it via the `enableFeatures`. I will also add it to the readme.

karthik2804:
I think this should wait until we get the [fetch PR](https://github.com/bytecodealliance/StarlingMonkey/pull/52), so leaving it as a draft.

README.md
```diff
@@ -130,6 +130,10 @@ Setting `disableFeatures: ['random', 'stdio', 'clocks']` will disable all featur
 
 Note that pure components **will not report errors and will instead trap**, so that this should only be enabled after very careful testing.
 
+The features that are not included by default are:
+* `'http'` - Support for sending and receiving HTTP requests, depends on `wasi:io` 
+* `'fetch-event'` - Enables using `fetchEvent` to respond to `wasi:http/incoming-handler@0.2.0#handle`. If the target world does note export `wasi:http/incoming-handler@0.2.0`, this will be ignored.
```
guybedford, line 135: Just like we automatically merge WASI imports into the target world, perhaps we can automatically merge the `wasi:http/incoming-handler@0.2.0` export into the target world when this feature is used?

guybedford: Actually isn't this the default behaviour anyway? Or do we still strip the export if it's not explicitly in the target world?

karthik2804: With this PR we want to only optionally strip the export from the StarlingMonkey engine [here](https://github.com/bytecodealliance/ComponentizeJS/blob/544b4edd28cb49d91d88b1abcd333712c65ecf91/crates/spidermonkey-embedding-splicer/src/splice.rs#L67-L74) based on the value of the feature. My intention with that statement is to make it such that, if the target world does not export the interface but the feature is enabled, we should probably still strip the export.

guybedford: Shouldn't it read "If the target world does export" though? Rather than "does not". Thinking about this further we probably want to make this an explicit error. That is, if you enable this feature and the target world already exports the incoming handler, then we should throw an error that you should either not target this export, or you should use the fetch event version of it, but not both.

karthik2804: I think there is a disconnect in what we each mean by the target world. I am talking about the `wit` definition, not the guest content. As in, consider the following world

```
package local:hello;

world hello {
  export hello: func(name: string) -> string;
}
```
and the `fetch-event` is set, if we do not strip the export from the engine, it will end up in the output component but maybe that i fine?  

I also think that there is currently no way to check if a handler is attached to the fetch event but we can check if there is something that targets the incoming handler export so erroring on that would be good, I agree.

guybedford: We are referring to the same thing. I'm saying that in the above, we do merge in the WASI world automatically even if it wasn't originally targeted. But that if it is explicitly targeted, the user might want to write their own custom handler something like the test case in https://github.com/bytecodealliance/jco/blob/main/test/fixtures/componentize/wasi-http-proxy/source.js#L102.

tschneidereit: I had written a long comment explaining why it felt too magical to me to have this option automatically add the `incoming-handler` export, but in writing it up I've come around and now think this is fine to do :)

The key insight is that we're talking about an export, not an import: it's valid to export interfaces/functions that aren't part of the target world. It means that those exports won't be called, but that's fine.

karthik2804: In case of writing an app that will be  used as a http trigger using the fetch event, wouldn't the host expect to call the export even though it is not part of the target world explicitly?

tschneidereit: If the developer is specifying a target world that is different from what their use case actually needs, and their runtime provides, then that's a bug in the configuration. With what Guy is proposing here, there'd be a higher likelihood of things sort of accidentally working out regardless. That could be seen as a downside, but I'm not even entirely sure it is, honestly.

karthik2804: Maybe I am misunderstanding here. To verify my understanding - the proposal above is to not retain `fetchEvent` incase the target world contains the `wasi:http/incoming-handler` export and only have the feature take effect if it is not explicitly part of the target world.

tschneidereit: Oh, I see—I had missed that part of it. On that, I agree with you: I think this setting should be usable whether the targeted world includes `incoming-handler` or not.

@guybedford doing what you proposed would make it substantially harder to use `FetchEvent`, because now anyone who wants to target an environment that supports `wasi:http/proxy` has to edit the target world not to include `incoming-handler` if they want to use `FetchEvent`. I think that's a requirement we really shouldn't put on people—and it'd be quite hard for us to satisfy using Spin as a development tool.

karthik2804: I am happy to update this PR once there is consensus on this discussion.

tschneidereit: Guy and I just discussed this further, and came to the realization that there's a cleaner way to do this: with #117, ComponentizeJS will only generate exports for symbols exported at the JS level, too. Based on that, we could simply say "if after treeshaking we still would export the symbol, then we'll do so. Otherwise, we'll leave it alone."

In effect, that'd mean that if the input StarlingMonkey already includes an export, and content doesn't specify that same export, ComponentizeJS will leave the original export alone.

@karthik2804, if that makes sense to you, then changing #117 accordingly should make it so that this issue isn't needed anymore.

karthik2804: I think I am roughly tracking it. Currently, #117 is purely for imports. I will take a stab at updating it and see if I understand things correctly and report back.

guybedford:
Please let me know if I can give any further feedback on this PR, or if you want to arrange a sync to discuss further @karthik2804.

karthik2804:
@guybedford I think this PR can be closed now as #117 enables the use of the fetch event handler as long as the guest content does not export the `incomingHandler` object.

karthik2804:
Checking back on this - this is not solved by #117 as we decided to remove changes affecting exports. Are we okay to still follow the plan described in https://github.com/bytecodealliance/ComponentizeJS/pull/116#discussion_r1658935176 but narrow it down to only the case of `wasi-http` export? or do we want to go down this route of adding it as a feature as is currently in this PR.

karthik2804:
The above is due to us exiting during the `exports_bindgen` call [here](https://github.com/bytecodealliance/ComponentizeJS/blob/498be912a6e0af1d360adc547b9ca119e4a15ac9/crates/spidermonkey-embedding-splicer/src/bindgen.rs#L383) and we do not even get to the point of optionally removing the export [here](https://github.com/bytecodealliance/ComponentizeJS/blob/498be912a6e0af1d360adc547b9ca119e4a15ac9/crates/spidermonkey-embedding-splicer/src/splice.rs#L53-L65)

tschneidereit:
I think we should still pursue the plan described in that comment, yes. Based on our conversation just now, that'll require parsing the input `.wasm` file earlier in the process so we can check whether it has the export in question. That seems okay to me, assuming it's not too challenging to do.

guybedford:
I haven't tested the workflows here, but the code path in https://github.com/bytecodealliance/ComponentizeJS/blob/main/crates/spidermonkey-embedding-splicer/src/bindgen.rs#L383 no longer does any skipping at all.

So we are just always leaving the incoming handler export in the final binary, even though it is not reflected in the encoded component.

Therefore it should be possible to simply target a world with the incoming handler to get it handled correctly in the component encoding. It should be possible to test this at the component level by stripping the component encoding and reencoding it with the exported hanlder.

Assuming I've got the above right, then the issue is that we don't use the guest exports to inform whether to use the JS handler or the engine handler when it is present in the world, and instead we will always assume a JS handler thereby overriding the engine one. The way to filter this would be to have an exception for this specific function in that codepath and I'd be fine with that.

karthik2804:
@guybedford I believe you are correct in that the JS handler always overrides the Engine handler. To make sure I understand, are you suggesting that we make an exception just for the HTTP handler specifically with something like [before this](
https://github.com/bytecodealliance/ComponentizeJS/blob/498be912a6e0af1d360adc547b9ca119e4a15ac9/crates/spidermonkey-embedding-splicer/src/bindgen.rs#L388)
```rust
if (name == "wasi:http/incoming-handler@0.2.0") {
  continue;
}
```

Alternatively, we could potentially parse the engine earlier and pass in the list of exports it is okay to ignore like @tschneidereit suggests.

I am happy with the first approach as well since that is simpler.  Happy to open a new PR once an approach is decided. Thanks!

guybedford:
Something like that sounds like it would work to me, yes.

karthik2804:
@guybedford do you have a preference between adding an exception for specific export or making it more general by parsing the engine before we get to the `export_bindgen`

guybedford:
Since this is just for guest exports (since imports always coalesce so there are no conflicts to deal with) and incoming HTTP is the only one currently, hard-coding should be fine.

guybedford:
Supported in https://github.com/bytecodealliance/ComponentizeJS/pull/122.

PR #122, opened 2024-07-18
allow  using fetchEvent for wasi:http/incoming-handler@0.2.0 optionally

karthik2804:
This PR adds the ability to use the `fetchEvent` from StarlingMonkey to handle incoming HTTP requests. This needs the `http` feature to be enabled explicitly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,10 @@ Note that pure components **will not report errors and will instead trap**, so t
 Note that features explicitly imported by the target world cannot be disabled - if you target a component to a world
 that imports `wasi:clocks`, then `disableFeatures: ['clocks']` will not be supported.
 
+## Using StarlingMonkey's `fetch-event`
+
+The StarlingMonkey engine provides the ability to use `fetchEvent` to handle calls to `wasi:http/incoming-handler@0.2.0#handle`. When targeting worlds that export `wasi:http/incoming-handler@0.2.0`, the component will use the `fetchEvent` if the guest content does not explicitly export a `incomingHandler` or `wasi:http/incoming-handler@0.2.0` object. Using the `fetchEvent` requires enabling the `http` feature. 
+
 ## API
 
 ```ts
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ that imports `wasi:clocks`, then `disableFeatures: ['clocks']` will not be suppo
 
 ## Using StarlingMonkey's `fetch-event`
 
-The StarlingMonkey engine provides the ability to use `fetchEvent` to handle calls to `wasi:http/incoming-handler@0.2.0#handle`. When targeting worlds that export wasi:http/incoming-handker@0.2.0 the fetch event will automatically be attached. Alternatively, to override the fetch event with a custom handler, export an explict incomingHandler or wasi:http/incoming-handler@0.2.0 object. Using the `fetchEvent` requires enabling the `http` feature. 
+The StarlingMonkey engine provides the ability to use `fetchEvent` to handle calls to `wasi:http/incoming-handler@0.2.0#handle`. When targeting worlds that export `wasi:http/incoming-handler@0.2.0` the fetch event will automatically be attached. Alternatively, to override the fetch event with a custom handler, export an explict `incomingHandler` or `'wasi:http/incoming-handler@0.2.0'` object. Using the `fetchEvent` requires enabling the `http` feature. 
 
 ## API
 
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ that imports `wasi:clocks`, then `disableFeatures: ['clocks']` will not be suppo
 
 ## Using StarlingMonkey's `fetch-event`
 
-The StarlingMonkey engine provides the ability to use `fetchEvent` to handle calls to `wasi:http/incoming-handler@0.2.0#handle`. When targeting worlds that export `wasi:http/incoming-handler@0.2.0`, the component will use the `fetchEvent` if the guest content does not explicitly export a `incomingHandler` or `wasi:http/incoming-handler@0.2.0` object. Using the `fetchEvent` requires enabling the `http` feature. 
+The StarlingMonkey engine provides the ability to use `fetchEvent` to handle calls to `wasi:http/incoming-handler@0.2.0#handle`. When targeting worlds that export wasi:http/incoming-handker@0.2.0 the fetch event will automatically be attached. Alternatively, to override the fetch event with a custom handler, export an explict incomingHandler or wasi:http/incoming-handler@0.2.0 object. Using the `fetchEvent` requires enabling the `http` feature. 
 
 ## API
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -133,6 +133,10 @@ Note that pure components **will not report errors and will instead trap**, so t
 Note that features explicitly imported by the target world cannot be disabled - if you target a component to a world
 that imports `wasi:clocks`, then `disableFeatures: ['clocks']` will not be supported.
 
+## Using StarlingMonkey's `fetch-event`
+
+The StarlingMonkey engine provides the ability to use `fetchEvent` to handle calls to `wasi:http/incoming-handler@0.2.0#handle`. When targeting worlds that export `wasi:http/incoming-handler@0.2.0`, the component will use the `fetchEvent` if the guest content does not explicitly export a `incomingHandler` or `wasi:http/incoming-handler@0.2.0` object. Using the `fetchEvent` requires enabling the `http` feature.
```
karthik2804, line 138: I would like some feedback on the wording and how we would like to document these changes.

guybedford: Looks good. Perhaps split the description into two separate sentences here:

> When targeting worlds that export `wasi:http/incoming-handker@0.2.0` the fetch event will automatically be attached. Alternatively, to override the fetch event with a custom handler, export an explict `incomingHandler` or `wasi:http/incoming-handler@0.2.0` object.

crates/spidermonkey-embedding-splicer/src/bindgen.rs
```diff
@@ -385,6 +385,11 @@ impl JsBindgen<'_> {
                         if let Some(name) = iface.name.as_ref() {
                             let camel_case_name = name.to_lower_camel_case();
                             if !guest_exports.contains(&camel_case_name) {
+                                if name == "incoming-handler"
```
guybedford, line 388: Let's just add a brief comment here as well to ensure this is properly motivated.

guybedford (approved):
Great, thanks for pushing on this. Happy to merge.

crates/spidermonkey-embedding-splicer/src/bindgen.rs
karthik2804: added.

README.md
```diff
@@ -133,6 +133,10 @@ Note that pure components **will not report errors and will instead trap**, so t
 Note that features explicitly imported by the target world cannot be disabled - if you target a component to a world
 that imports `wasi:clocks`, then `disableFeatures: ['clocks']` will not be supported.
 
+## Using StarlingMonkey's `fetch-event`
+
+The StarlingMonkey engine provides the ability to use `fetchEvent` to handle calls to `wasi:http/incoming-handler@0.2.0#handle`. When targeting worlds that export wasi:http/incoming-handker@0.2.0 the fetch event will automatically be attached. Alternatively, to override the fetch event with a custom handler, export an explict incomingHandler or wasi:http/incoming-handler@0.2.0 object. Using the `fetchEvent` requires enabling the `http` feature.
```
guybedford, line 138: Minor adjustments:

```suggestion
The StarlingMonkey engine provides the ability to use `fetchEvent` to handle calls to `wasi:http/incoming-handler@0.2.0#handle`. When targeting worlds that export `wasi:http/incoming-handler@0.2.0` the fetch event will automatically be attached. Alternatively, to override the fetch event with a custom handler, export an explict `incomingHandler` or `'wasi:http/incoming-handler@0.2.0'` object. Using the `fetchEvent` requires enabling the `http` feature. 
```

I'm just wondering about the `http` feature requirement. I suppose we will disable this once fetch is stable, but even then we need to think about what to do when `http` is disabled.

We should probably make it an error when http is disabled and the user hasn't exported the handler. Either just the error already at https://github.com/bytecodealliance/ComponentizeJS/pull/122/files#diff-69c4ddcc3664701394605dc159a10360361dc21afeee1a01c967c1f06f1f967aR398, or a new error that specifies the feature must be enabled or exported.

karthik2804: I will pipe it down to the `bindgen` and create a new error.

karthik2804: updated.

guybedford (commented):
Finally I think the guest exports check should actually check both the camel name and the full interface name, so:

```rs
if !guest_exports.contains(&camel_case_name) && !guest_exports.contains(name) {
```

But looking at this again, it seems `camel_case_name` isn't quite right, as it should probably splice off the `:` and `/` parts.

This makes me wonder if this PR is actually working correctly? Can you confirm?

karthik2804:
I tested this only with the `incomingHandler` annotation but the above is a good catch. Let me go do some more testing/digging and report back.

guybedford:
It might be that `name` is just the interface name and so that the check should be using `name_world_key` against the interface world key for the full `wasi:incoming-handler@2.0.0` case.

guybedford:
Ahh, `name` is shadowing the outer name, so it should be renamed to `Some(iface_name) = ` I think.

karthik2804:
To confirm my understanding the allowed exports are  `incomingHandler` and `"wasi:http/incoming-handler@0.2.0"`  correct? or is the explicit notation supposed to be `wasiHttpIncomingHandler020` ?

guybedford:
Yes exactly, so checking `iface_name.toCamelCase()` and `name` which are separate originating values.

karthik2804:
I believe we are already doing both the checks on line 388 (fully qualified name) and line 392 (sugared name) unless I am missing something.

crates/spidermonkey-embedding-splicer/src/bindgen.rs
```diff
@@ -382,9 +387,25 @@ impl JsBindgen<'_> {
...
+                        if let Some(iface_name) = iface.name.as_ref() {
+                            let camel_case_name = iface_name.to_lower_camel_case();
                             if !guest_exports.contains(&camel_case_name) {
+                                // For wasi:http/incoming-handler, we treat it
+                                // as a special case as the engine already
+                                // provides the export using fetchEvent and that
+                                // can be used when an explicit export is not
+                                // defined by the guest content.
+                                if iface_name == "incoming-handler"
+                                    || iface_name == "wasi:http/incoming-handler@0.2.0"
```
karthik2804, line 399: `iface_name` here will always be `incoming-handler` correct? or is there a case where it will be fully qualified?

guybedford: Right, so we should be checking I think:

```suggestion
                                if iface_name == "incoming-handler"
                                    || name == "wasi:http/incoming-handler@0.2.0"
```

right?

karthik2804: Yep that sounds correct to me. I'll push the changes once I'm back.

karthik2804: updated.

karthik2804:
@guybedford let me know if there is anything else to address for this PR.

karthik2804:
@guybedford would it be possible to get a release with this PR? Thanks!

guybedford:
Sure, I've published a 0.10.1.

PR #134, opened 2024-09-04
add option to enable weval

### LINKED ISSUE #133: Support building with `weval`
I think it would be nice to optionally allow users to build components that are `weval`'d to get the performance boosts that it provides. I did manage to get a build locally working with `weval` but I am not sure on how we would like to package it in `ComponentizeJS`. 

We could include the build of StarlingMonkey for both cases with and without `weval` and also the ics-cache. This would increase the size of the package. We could optionally look at adding an optional dependency containing the build artifacts for aot if we prefer not to increase the size of the `ComponentizeJS` package. 

I am happy to open a PR once a preferred approach is identified.

karthik2804:
closes #133 

I am not sure how we want to modify the tests, Do we want to run all the tests for the `weval` build as well? I am happy to update it as necessary.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,8 +21,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
-    runs-on: ${{ matrix.os }}
+        config:
+          - { os: ubuntu-latest, build: "npm run build", test: "npm run test", cacheDirectory: "build-release", cacheKeySuffix: ""  }
+          - { os: ubuntu-latest, build: "npm run build:weval", test: "npm run test:weval", cacheDirectory: "build-release-weval", cacheKeySuffix: "-weval" }
+    runs-on: ${{ matrix.config.os }}
     steps:
     - uses: actions/checkout@v2
       with:
@@ -56,18 +58,14 @@ jobs:
       uses: actions/cache@v3
       id: starlingmonkey-build
       with:
-        path: |
-          build-release
-          build-release-weval 
-        key: engine-build-${{ steps.starlingmonkey-commit.outputs.STARLINGMONKEY_HASH }}
+        path: ${{matrix.config.cacheDirectory}} 
+        key: engine-build${{matrix.config.cacheKeySuffix}}-${{ steps.starlingmonkey-commit.outputs.STARLINGMONKEY_HASH }}
 
     - name: Build
-      run: |
-        npm run build
-        npm run build:weval
-
+      run: ${{matrix.config.build}}
+    
     - name: Test
-      run: npm run test
+      run: ${{matrix.config.test}}
 
     - name: Cache Example build
       uses: actions/cache@v3
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,11 +56,15 @@ jobs:
       uses: actions/cache@v3
       id: starlingmonkey-build
       with:
-        path: build-release
+        path: |
+          build-release
+          build-release-weval 
         key: engine-build-${{ steps.starlingmonkey-commit.outputs.STARLINGMONKEY_HASH }}
 
     - name: Build
-      run: npm run build
+      run: |
+        npm run build
+        npm run build:weval
 
     - name: Test
       run: npm run test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,8 +22,8 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - { os: ubuntu-latest, build: "npm run build", test: "npm run test", cacheDirectory: "build-release", cacheKeySuffix: ""  }
-          - { os: ubuntu-latest, build: "npm run build:weval", test: "npm run test:weval", cacheDirectory: "build-release-weval", cacheKeySuffix: "-weval" }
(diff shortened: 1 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -56,11 +56,15 @@ jobs:
...
+        path: |
+          build-release
+          build-release-weval 
         key: engine-build-${{ steps.starlingmonkey-commit.outputs.STARLINGMONKEY_HASH }}
 
     - name: Build
-      run: npm run build
+      run: |
+        npm run build
+        npm run build:weval
```
guybedford, line 67: I think the best approach here is to use the matrix on both the test and build steps, to then build both versions in parallel and test both versions in parallel.

guybedford (approved):
Amazing, thank you!

For CI, building and testing both would be ideal.

karthik2804:
I am not sure why the tests for the weval build fail in the way they are failing. The generated `.d.ts` files from the transpile step are different. Adding only the difference in the types 

regular build
```ts
export function run(): void;
```

`weval` build 
```ts
import { WasiCliRun } from './interfaces/wasi-cli-run.js';
import { WasiHttpIncomingHandler } from './interfaces/wasi-http-incoming-handler.js';
export const run: typeof WasiCliRun;
export const incomingHandler: typeof WasiHttpIncomingHandler;
```

I will try to investigate this more.

src/componentize.js
```diff
@@ -193,7 +193,7 @@ export async function componentize(jsSource, witWorld, opts) {
           'weval',
           `--cache-ro ${aotCache}`,
           '-w',
-          `-i ${engine}`,
+          `-i ${input}`,
           `-o ${output}`,
           `--dir ${dirname(input)}`
```
guybedford, line 198: I believe all of the arguments to Weval should be exactly the same as the arguments to Wizer, as this was how Weval was designed. So I'd actually make sure all the flags match with the Wizer lines below here.

karthik2804:
@guybedford I believe I have the same arguments being passed to both but I still run into the error. I feel like I am missing something obvious.

src/componentize.js
```diff
@@ -181,42 +183,80 @@ export async function componentize(jsSource, witWorld, opts) {
...
+      throw new Error(err);
+    }
+  } else {
+    try {
+      let wizerProcess = spawnSync(
+        wizer,
+        [
+          '--allow-wasi',
+          '--init-func',
+          'componentize.wizer',
```
guybedford, line 230: `--init-func` must be provided.

src/componentize.js
```diff
@@ -181,42 +183,80 @@ export async function componentize(jsSource, witWorld, opts) {
...
-    } else {
-      rmSync(tmpDir, { recursive: true });
+      throw new Error(err);
+    }
+  } else {
+    try {
+      let wizerProcess = spawnSync(
+        wizer,
+        [
+          '--allow-wasi',
```
guybedford, line 228: `--allow-wasi` must be enabled I believe as well.

src/componentize.js
karthik2804:
```
error: Found argument '--init-func' which wasn't expected, or isn't valid in this context
```

These are not arguments that are passed in the reference I was using from the [fastly js-compute-runtime diff](https://github.com/fastly/js-compute-runtime/pull/883/files)

src/componentize.js
karthik2804: Looks like this. is not allowed as well.

```
error: Found argument '--allow-wasi' which wasn't expected, or isn't valid in this context
```

src/componentize.js
guybedford: `--init-func` is a Wizer option ComponentizeJS needs to use, and Weval accepts common Wizer options.

src/componentize.js
guybedford: Apparently this is on-by-default in Weval.

src/componentize.js
guybedford: Ahh Weval doesn't support this so we need to either work around this custom init function by using the common Wizer name here (updating `engine.cpp` to use `wizer.initialize`), or we need to get this option supported in Weval.

karthik2804: Would it be a possibility to run `wizer` first then followed by running `weval` without `-w` flag

guybedford: It should be relatively straightforward to add this option to Weval and pipe it through to Wizer here - https://github.com/cfallin/weval/blob/main/src/main.rs#L94.

Would you be able to look into this for Weval further?

karthik2804: Yep - I just started looking at it. Will open a PR soon.

guybedford: Chris is already on it - https://github.com/cfallin/weval/commit/81226fd1a23ab446be31330d5fb55ec387ffe101, so next Weval release should be good to support this.

karthik2804: Oh, that was fast! Thanks!

guybedford: This should be in Weval 0.2.13 now.

karthik2804: Just pushed the changes and the test seems to pass locally so far!

guybedford:
The failing test remaining seems like it is due to Weval taking slightly longer to build.

Updating the test time diff from 5 seconds to 10 or 15 seconds like the following should fix:

```js
  // verify now was taken at build time (within than 10 seconds ago)
  ok(Number(times[0]) < curNow);
  ok(Number(times[0]) > curNow - 10_000);
}
```

karthik2804:
@guybedford thank you for the final test fix and helping get this PR over the line!

PR #136, opened 2024-09-16
allow to optionally pass in host env to wizer/weval

### LINKED ISSUE #135: Host Env variables are not passed through to `wizer`
We currently override the envs that we provide to the `wizer` process.  [here](https://github.com/bytecodealliance/ComponentizeJS/blob/5c06448136fd8c6852b60b3a5f4c8c2f188e6734/src/componentize.js#L161-L166). This leads to `nix` like systems not being able to run `ComponentizeJS` as it needs some `NIX_LD_*` variables set. 

Is there a reason to limit the env var available to the `wizer` process or should we be okay with appending the host env vars  to the existing set.

karthik2804:
closes #135

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -43,6 +43,12 @@ interface ComponentizeOptions {
    * (no experimental subsystems currently supported)
    */
   enableFeatures?: [],
+  /**
+   * Pass environment variables to the spawned Wizer or Weval Process If set to
+   * true, all host environment variables are passed. To pass only a subset,
+   * provide an object with the desired variables.   
+   */
+  env: boolean | Record<string, string>,
 }
 
 /**
--- a/src/componentize.js
+++ b/src/componentize.js
@@ -158,7 +158,16 @@ export async function componentize(jsSource, witWorld, opts) {
     )
   );
 
+  let hostenv = {};
+
+  if (opts.env) {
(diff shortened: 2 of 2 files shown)


types.d.ts
```diff
@@ -43,6 +43,12 @@ interface ComponentizeOptions {
    * (no experimental subsystems currently supported)
    */
   enableFeatures?: [],
+  /**
+   * Pass environment variables to the spawned Wizer or Weval Process If set to
+   * true, all host environment variables are passed. To pass only a subset,
+   * provide an object with the desired variables.   
+   */
+  passthroughHostEnv: boolean | Record<string, string>,
```
guybedford, line 51: `passthrough` usually means you're passing something directly through from one thing to another, which isn't the case in case of the second type in the union here. `passThrough` as two words might be clearer on that. But maybe let's just call this `hostEnv`?

Also wouldn't `guestEnv` be more correct than `hostEnv`? Or even just `env`?

guybedford (approved):
Looks great, thanks. Just some naming questions.

types.d.ts
karthik2804: I will go with just `env`. That sounds cleaner.

karthik2804: updated.

Issue #144, opened 2024-10-18
Set performance time origin


guybedford:
We should be setting this for ComponentizeJS at runtime.

Runtime start is effectively the call to the first export, so we can set this if not already set in the export call handler.

guybedford:
Also when not enabling the `clocks` feature, we should turn this off entirely.

PR #146, opened 2024-10-22
set timeOrigin in componentize_initialize

cataggar:
I attempted #144, but I was not able to go back and reproduce the failure.

### DIFF
--- a/embedding/embedding.cpp
+++ b/embedding/embedding.cpp
@@ -1,4 +1,5 @@
 #include "embedding.h"
+#include "builtins/web/performance.h"
 
 namespace builtins::web::console {
 
@@ -350,6 +351,7 @@ componentize_initialize() {
     Runtime.debug = true;
   }
 
+  builtins::web::performance::Performance::timeOrigin = std::chrono::steady_clock::now();
   __wizer_initialize();
   char env_name[100];
   LOG("(wizer) retrieve and generate the export bindings");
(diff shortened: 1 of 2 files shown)


embedding/embedding.cpp
```diff
@@ -350,6 +351,7 @@ componentize_initialize() {
     Runtime.debug = true;
   }
 
+  builtins::web::performance::Performance::timeOrigin = std::chrono::steady_clock::now();
```
guybedford, line 354: Instead of in initialize, this should be done on the call to the first export, ie in the `__attribute__((export_name("call"))) uint32_t call` function, but with a check that it has not already been done.

guybedford (commented):
Thanks for giving this a go, it's not too far off the mark.

cataggar:
Thanks. I moved it as requested. I'm not really sure how to add a unit test for this.

embedding/embedding.cpp
```diff
@@ -145,6 +146,10 @@ cabi_realloc(void *ptr, size_t orig_size, size_t org_align, size_t new_size) {
 
 __attribute__((export_name("call"))) uint32_t call(uint32_t fn_idx,
                                                    void *argptr) {
+  if (!builtins::web::performance::Performance::timeOrigin.has_value()) {
```
cataggar, line 149: If `timeOrigin` is not set, perhaps this does not work. Tests are breaking.

cataggar:
Sorry, I don't know what I'm doing.

guybedford:
You were close! https://github.com/bytecodealliance/ComponentizeJS/pull/150

Issue #145, opened 2024-10-17
Resource function names should be kebab-cased


cataggar:
I've been trying to make my first resource definition. I keep getting an error message that my `compilerOptions` resource function is not valid.

```
$ bunx jco componentize cowsay.js --wit src/cowsay.wit -o cowsay.wasm
(jco componentize) ComponentError: failed to encode a component from module
$failed to decode world from module

Caused by:
    0: module was not valid
    1: import of `[resource-new]compilerOptions` is not a valid resource function
    at componentNew (file:///Users/cataggar/ms/jco/obj/wasm-tools.js:3714:11)
    at componentNew (file:///Users/cataggar/ms/jco/src/api.js:37:10)
    at async componentize (file:///Users/cataggar/ms/ComponentizeJS/src/componentize.js:381:5)
    at async componentize (file:///Users/cataggar/ms/jco/src/cmd/componentize.js:11:25)
    at async file:///Users/cataggar/ms/jco/src/jco.js:200:9
error: script "componentize" exited with code 1
```

Here is my cowsay.wit where I'm trying to define an empty `resource compiler-options`:
```wit
package local:cowsay;

world cowsay {
  export test: func() -> string;
  export canvas: interface {
    ts-version: func() -> string;
    resource compiler-options {
      // constructor();
    }
  }
}
```

Here is my cowsay.ts where I'm trying to define it:
``` ts
import * as ts from 'typescript';
export function test() {
    return "ok";
}

class CompilerOptions {
    inner;
    constructor() {
        this.inner = 3;
    }
}

function tsVersion() {
    return ts.version;
}

export const canvas = {
    tsVersion: tsVersion,
    compilerOptions(){
        return new CompilerOptions();
    }
    // compilerOptions: (): CompilerOptions => {
    //     return new CompilerOptions();
    // }
};
```

guybedford:
Thanks for confirming, filed as a ComponentizeJS bug.

cataggar:
I was reviewing code this evening. Is `crates/spidermonkey-embedding-splicer/src/bindgen.rs` the right place for this functionality?

guybedford:
It should be, yes.

cataggar:
I wasn't unable to hunt this one down. Would it help if I attached the generated sources?

guybedford:
Fix in https://github.com/bytecodealliance/ComponentizeJS/pull/151.

PR #151, opened 2024-10-29
fix: ensure resource imports and exports use kebab case

guybedford:
This fixes a bug where resource imports and exports were not using kebab case in the component model ABI, resulting in failures for resource names of more than one word.

Resolves https://github.com/bytecodealliance/ComponentizeJS/issues/145.


cataggar:
Thank you! This unblocks me.

Issue #153, opened 2024-10-30
timeOrigin still not set


cataggar:
I still get the same error with jco 1.7.1 that uses 0.13.1 of this project. #150 does not set timeOrigin. The change from https://github.com/bytecodealliance/ComponentizeJS/pull/146/commits/cb4d73af4fbbd75fb4bafc79b04a7d500180bd96#diff-882dd8ad95be11b373fd2693a697c12144f8b51d4605a35cd678a0d2e5b4a09dR354 makes it work for me locally.

```
~/ms/StarlingMonkey15/cowsay> npx jco componentize cowsay.js --wit src/cowsay.wit -o cowsay.wasm
Redirecting call to abort() to mozalloc_abort

Error: the `componentize.wizer` function trapped

Caused by:
    0: error while executing at wasm backtrace:
           0: 0x25f20 - <unknown>!<wasm function 206>
           1: 0x25f29 - <unknown>!<wasm function 207>
           2: 0xa921 - <unknown>!<wasm function 87>
           3: 0x42119 - <unknown>!<wasm function 348>
           4: 0x25431f - <unknown>!<wasm function 5279>
           5: 0x254f50 - <unknown>!<wasm function 5282>
           6: 0x25576e - <unknown>!<wasm function 5285>
           7: 0x35a029 - <unknown>!<wasm function 7091>
           8: 0x258ee8 - <unknown>!<wasm function 5313>
           9: 0x24bbcc - <unknown>!<wasm function 5278>
          10: 0x245b22 - <unknown>!<wasm function 5276>
          11: 0x255aeb - <unknown>!<wasm function 5288>
          12: 0x29a0a9 - <unknown>!<wasm function 5791>
          13: 0x34e9db - <unknown>!<wasm function 7049>
          14: 0x34e72d - <unknown>!<wasm function 7049>
          15: 0x20286 - <unknown>!<wasm function 105>
          16: 0x1f50e - <unknown>!<wasm function 104>
          17: 0xa882 - <unknown>!<wasm function 86>
          18: 0x22580f - <unknown>!<wasm function 5012>
    1: wasm trap: wasm `unreachable` instruction executed
(jco componentize) Error: Failed to initialize the compiled Wasm binary with Wizer:
Wizering failed to complete
    at componentize (file:///Users/cataggar/ms/StarlingMonkey15/node_modules/@bytecodealliance/componentize-js/src/componentize.js:271:13)
[...]

PR #154, opened 2024-10-30
set timeOrigin in componentize_initialize

cataggar:
This workaround still works for me for #153. Same as first commit from #146.


guybedford:
Capturing time origin at Wizer time is not a feature we support.

PR #155, opened 2024-11-07
Handle undefined `opts` and allow undefined `env`

primoly:
[This example](https://github.com/bytecodealliance/ComponentizeJS/blob/main/EXAMPLE.md) doesn’t currently work as the the code in [componentize.js](https://github.com/bytecodealliance/ComponentizeJS/blob/main/src/componentize.js) fails on this line https://github.com/bytecodealliance/ComponentizeJS/blob/d4ae59c16c0c5a781f3c4c18a4410a5fcf4e73a0/src/componentize.js#L50 because it can’t handle an undefined `opts`, even tho this should be allowed as suggested here https://github.com/bytecodealliance/ComponentizeJS/blob/d4ae59c16c0c5a781f3c4c18a4410a5fcf4e73a0/types.d.ts#L65 This PR should fix this.

I also allow the `env` field of `ComponentizeOptions` to be optional like all the others.

### DIFF
--- a/src/componentize.js
+++ b/src/componentize.js
@@ -37,6 +37,7 @@ export async function componentize(jsSource, witWorld, opts) {
     opts = witWorld;
     witWorld = opts?.witWorld;
   }
+  opts = opts || {};
   const {
     sourceName = 'source.js',
     preview2Adapter = preview1AdapterReactorPath(),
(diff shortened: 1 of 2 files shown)


guybedford:
Thank you for the PR, will land as soon as CI is green.

guybedford (approved):

PR #161, opened 2024-11-21
fix: typescript type of package

### LINKED ISSUE #160: Types seems to be wrongly exported and also not included in published package
I recently started using this package and tried running the example code. However, my linter flagged an issue: there are no type declarations available.

It appears that the `types.d.ts` file is not included in the published package. This can be verified by inspecting the package directly on the NPM page: https://www.npmjs.com/package/@bytecodealliance/componentize-js?activeTab=code

---

Additionally, it seems there's another issue related to how the types are handled even if the types has been properly shipped alongside the package.

Since the package uses the `exports` field in `package.json`, the `types` will not be solved. To address this, seems the package should either:

- Set the `types` field in the `exports` object in `package.json`.
- Use the `module` field instead of `exports` to make the `types` accessible.

Here's the specific error message from TypeScript:

```
Could not find a declaration file for module '@bytecodealliance/componentize-js'. './node_modules/.pnpm/@bytecodealliance+componentize-js@0.14.0/node_modules/@bytecodealliance/componentize-js/src/componentize.js' implicitly has an 'any' type.
  There are types at './node_modules/@bytecodealliance/componentize-js/types.d.ts', but this result could not be resolved when respecting package.json "exports". The '@bytecodealliance/componentize-js' library may need to update its package.json or typings. ts(7016)
```

moontai0724:
- Added type declaration files to ensure they are shipped alongside the package.
- Adjusted the exports field in package.json to ensure TypeScript can properly resolve the types.

Fixes #160

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,10 @@
   "homepage": "https://github.com/bytecodealliance/componentize-js#readme",
   "description": "ESM -> WebAssembly Component creator, via a SpiderMonkey JS engine embedding",
   "type": "module",
-  "exports": "./src/componentize.js",
+  "exports": {
+    "types": "./types.d.ts",
+    "import": "./src/componentize.js"
+  },
   "devDependencies": {
     "@bytecodealliance/preview2-shim": "^0.16.7",
     "mocha": "^10.2.0"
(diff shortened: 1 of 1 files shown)


guybedford:
Thank you!

Issue #164, opened 2024-11-29
utf8Encode failing on undefined


ctaggart:
```
~/ms/StarlingMonkey15/cowsay> bunx jco componentize cowsay.js --wit ../wit -o cowsay.wasm
(jco componentize) TypeError: expected a string, not type: undefined
    at utf8Encode (file:///Users/cataggar/ms/StarlingMonkey15/node_modules/@bytecodealliance/componentize-js/lib/spidermonkey-embedding-splicer.js:118:11)
    at spliceBindings (file:///Users/cataggar/ms/StarlingMonkey15/node_modules/@bytecodealliance/componentize-js/lib/spidermonkey-embedding-splicer.js:3760:47)
    at componentize (file:///Users/cataggar/ms/StarlingMonkey15/node_modules/@bytecodealliance/componentize-js/src/componentize.js:87:64)
    at async componentize (file:///Users/cataggar/ms/StarlingMonkey15/node_modules/@bytecodealliance/jco/src/cmd/componentize.js:11:25)
    at async file:///Users/cataggar/ms/StarlingMonkey15/node_modules/@bytecodealliance/jco/src/jco.js:200:9
```

The function that is throwing is:

``` js
function utf8Encode(s, realloc, memory) {
  if (typeof s !== 'string') throw new TypeError('expected a string');
  if (s.length === 0) {
    utf8EncodedLen = 0;
    return 1;
  }
  let buf = utf8Encoder.encode(s);
  let ptr = realloc(0, 0, 1, buf.length);
  new Uint8Array(memory.buffer).set(buf, ptr);
  utf8EncodedLen = buf.length;
  return ptr;
}
```

If I hack it to return if undefined with this code:

``` js
function utf8Encode(s, realloc, memory) {
  if (!s){
    utf8EncodedLen = 0;
    return 1;
  }
  if (typeof s !== 'string'){
    throw new TypeError(`expected a string, not type: ${typeof s}`);
  }
  if (s.length === 0) {
    utf8EncodedLen = 0;
    return 1;
  }
  let buf = utf8Encoder.encode(s);
  let ptr = realloc(0, 0, 1, buf.length);
  new Uint8Array(memory.buffer).set(buf, ptr);
  utf8EncodedLen = buf.length;
  return ptr;
}
```

Then I get this error:

```
~/ms/StarlingMonkey15/cowsay> bunx jco componentize cowsay.js --wit ../wit -o cowsay.wasm
[...]

PR #167, opened 2024-12-11
fix: specifier mappings in ComponentizeJS

guybedford:
This fixes an issue where we would incorrectly try to map `import.meta.url` due to not checking the import type.

Resolves https://github.com/bytecodealliance/ComponentizeJS/issues/164.


guybedford:
I've also pushed up a fix here to provide better error messages for incorrect imports finally.

cataggar:
I’m looking forward to trying it. Thanks!

PR #172, opened 2024-12-17
On Windows, do not require redist of vcruntime DLL

itowlson:
The Rancher Desktop folks are looking to bundle some Wasm developer tooling. However, they cannot redistribute the vcruntime DLL on Windows for licensing reasons, and therefore prefer Windows builds in the toolchain to link the CRT statically instead of dynamically.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[target.'cfg(all(windows, target_env = "msvc"))']
+rustflags = ["-C", "target-feature=+crt-static"]


guybedford:
Happy to accommodate although ComponentizeJS only builds for Wasm currently - can you share some further info on how this works? I'd be uncomfortable landing a build config whose target is not tested in this repo, without understanding more about the use case.

itowlson:
Ah, sorry, this is down to me making an assumption and not checking it with more knowledgeable folks - for some reason I thought ComponentizeJS was a native binary.  Closing and sorry for the noise!

PR #174, opened 2024-12-24
fix: invalid usage example in README

hyp3rflow:
It seems minor but I think it would be great if just copy-and-paste works.


guybedford:
Thanks for sharing.

PR #178, opened 2025-01-16
Allow to retain fetchEvent on all wasi 0.2.x versions

karthik2804:
Fixes the case where the guest wit states `wasi:http/incoming-handler@0.2.0` but the engine exports `wasi:http/incoming-handler@0.2.2`.

### DIFF
--- a/crates/spidermonkey-embedding-splicer/src/bindgen.rs
+++ b/crates/spidermonkey-embedding-splicer/src/bindgen.rs
@@ -429,7 +429,7 @@ impl JsBindgen<'_> {
                                 // can be used when an explicit export is not
                                 // defined by the guest content.
                                 if iface_name == "incoming-handler"
-                                    || name.starts_with("wasi:http/incoming-handler@0.2")
+                                    || name.starts_with("wasi:http/incoming-handler@0.2.")
                                 {
                                     if !features.contains(&Features::Http) {
                                         bail!(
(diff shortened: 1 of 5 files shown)


karthik2804:
Are we okay bumping this to 4s? I see the last few commits have also failed on it.
https://github.com/bytecodealliance/ComponentizeJS/blob/d8dd99536d148dc4afb3e4ae5f82ec4df92fb6d3/test/builtins/performance.js#L24-L27

crates/spidermonkey-embedding-splicer/src/bindgen.rs
```diff
@@ -429,7 +429,7 @@ impl JsBindgen<'_> {
                                 // can be used when an explicit export is not
                                 // defined by the guest content.
                                 if iface_name == "incoming-handler"
-                                    || name == "wasi:http/incoming-handler@0.2.0"
+                                    || name.starts_with("wasi:http/incoming-handler@0.2")
```
guybedford, line 432: This would also match `0.20` - I think we need "0.2." as the prefix.

guybedford (changes requested):
Yes, bumping CI to 4 or 5 seconds would be fine.

crates/spidermonkey-embedding-splicer/src/bindgen.rs
karthik2804: ah yeah - Thanks! updated.

guybedford (approved):

crates/spidermonkey-embedding-splicer/src/bindgen.rs
pchickey: You pushed a fix in the other occurrence but not at this spot

karthik2804: oh my bad - I forgot to save the file. fixed it now

PR #179, opened 2025-01-16
deps: update to latest StarlingMonkey

karthik2804:
Updates StarlingMonkey to [257fece3b178385f2ec898008ccb951b52e5fce1](https://github.com/bytecodealliance/StarlingMonkey/tree/257fece3b178385f2ec898008ccb951b52e5fce1)


guybedford:
I've just landed https://github.com/bytecodealliance/StarlingMonkey/pull/187 as well which would be great to include here.

karthik2804:
Will update the PR.

karthik2804:
@guybedford would be possible to get a new release? Thanks! I am happy to send a bump PR. I assume a patch release would work?

guybedford:
@karthik2804 ideally we should do a full WASI upgrade first I think, including https://github.com/bytecodealliance/StarlingMonkey/pull/174 to the latest version as well. Or do you need this very urgently?

tschneidereit:
@guybedford can you say more about why you want to block a release on a WASI upgrade? AFAICT, we'd not actually make use of any of the new functionality provided by the new WASI version, right?

guybedford:
@tschneidereit ComponentizeJS will merge the world imports with the componentized imports, which will result in components importing old WASI versions as well as the new WASI versions for the component being made, without merging, which should be avoided due to import bloat if possible. Not blocking the release just saying I would prioritize avoiding that for a release myself.

PR #180, opened 2025-01-21
bump StarlingMonkey to include wasi 0.2.3

karthik2804:
Updates StarlingMonkey dependency to [5ed4e1faf4b2a6cd3e2cbde154eb064e77c1c991](https://github.com/bytecodealliance/StarlingMonkey/commit/5ed4e1faf4b2a6cd3e2cbde154eb064e77c1c991).


guybedford:
Thanks for sharing! We also need to run the `update-wasi.sh` script in the root of the repo to complete updating ComponentizeJS to this version.

karthik2804:
It seems to add `wasmtime` as a dependency is that to be expected and should it be committed/ added as submodule? Thanks!

guybedford:
Yes, it clones the submodule to copy the files, but don't check it in.

guybedford:
`test/wit/test1.wit` also needs to be updated per the CI error. Just let me know if you need further help.

PR #184, opened 2025-01-31
Fix componentize::GetMemBuffer

tschneidereit:
This patch fixes an issue where if the return value of `sbrk(0)` ever changed after initially being set, we'd encounter a panic: the value is stored in a `PersistentRootedObject`, `AB`, which the code tries to initialize multiple times. Persistent roots can't only be initialized ones, after which they have to be `set`.

### DIFF
--- a/embedding/embedding.cpp
+++ b/embedding/embedding.cpp
@@ -490,7 +490,7 @@ bool ReportAndClearException(JSContext *cx) {
   return true;
 }
 
-void *LAST_SBRK;
+void *LAST_SBRK = nullptr;
 JS::PersistentRootedObject AB;
 static bool GetMemBuffer(JSContext *cx, unsigned argc, JS::Value *vp) {
   if (sbrk(0) != LAST_SBRK) {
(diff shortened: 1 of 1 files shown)


tschneidereit:
I don't know how to build a reasonable test for this, but it fixes an issue we encountered in the real world.

guybedford:
Nice, I can imagine this must have been very tricky to track down. Thanks I wasn't aware of this persistent rooted behaviour!

guybedford:
Happy to merge, but for testing, what conditions would trigger it? Large memory use at Wizer time?

guybedford:
If so maybe it's as simple as a test with a large source text?

guybedford (approved):

tschneidereit:
It's happening at runtime, not during wizening. I think it might be related to malloc, not GC allocation, but that's hard to reduce

guybedford:
Right, so maybe a call with a very large buffer that causes a page size increase, followed by another call?

PR #187, opened 2025-02-21
Support passing arguments to the StarlingMonkey runtime

tschneidereit:
StarlingMonkey supports a growing number of arguments in addition to the filename of the script to run. This commit adds support for passing those arguments through when componentizing.

I think it'll eventually make sense to expose some of these arguments more explicitly, but providing this option allows us to evolve the interfaces separately, similar to how e.g. clang supports passing linker arguments, or cargo supports passing rustc arguments.

### DIFF
--- a/src/componentize.js
+++ b/src/componentize.js
@@ -50,6 +50,7 @@ export async function componentize(jsSource, witWorld, opts) {
       new URL(`../lib/starlingmonkey_ics.wevalcache`, import.meta.url)
     ),
   } = opts;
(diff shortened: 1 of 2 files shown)


tschneidereit:
The reason I'm opening this PR now is that this will enable us to use the [source debugging feature](https://github.com/bytecodealliance/StarlingMonkey/pull/218) we just landed in StarlingMonkey without a custom version of ComponentizeJS

guybedford:
Just updated the types.d.ts and docs files as well here. This approach looks great thanks.

PR #197, opened 2025-03-06
Support all wasi-0.2.x patch versions in the splicer

tschneidereit:
This way, we don't have to update ComponentizeJS for any future minor versions.


guybedford:
Makes a lot more sense, thanks. We should probably do something similar in WASI Virt as well.

PR #203, opened 2025-03-19
refactor: componentize changes

vados-cosmonic:
This refactoring makes a few changes to the code to make componentize-js a bit easier to debug:

- enable passing debugBindings, debugBinary, and debugCalls options
- enable passing where to output debug output
- split out function for easier reading


t3hmrman:
Setting this to draft while https://github.com/bytecodealliance/ComponentizeJS/pull/202 lands -- the functionality in here will have to change quite a bit.

src/componentize.js
```diff
@@ -1,64 +1,82 @@
...
+
+/**
+ * Create a WebAssembly component from the given JS source code
+ *
+ * @param {string} jsSource - Javascript source code
+ * @param {string} [witWorld] - WIT world that should be targeted
+ * @param {object} [opts] - Options for creating the component
+ * @param {object} [opts.debugBuild] - Whether to use a debug build of StarlingMonkey
+ * @param {object} [opts.debug] - Options for debugging
+ */
```
tschneidereit, line 74: The doc comment doesn't reflect the latest interface anymore.

What's more, I think we should leave the doc comment out here entirely: it'll be picked up from the `.d.ts` file, and we shouldn't have different docs in two places.
```suggestion
```

tschneidereit (commented):
Would you by any chance be able to split this into two commits, where one only does pure movement of code that's otherwise unmodified, and the other makes whatever functional or structural changes are contained?

I started reviewing this, but found myself jumping around and trying to follow where there are real changes, vs where a function has just moved to a different location in the file, and found that very hard to do.

tschneidereit:
(I appreciate that my last big PR to this repo did a lot of the same. I tried minimizing code movement in that, but in hindsight should nevertheless have split the movement that I did apply off into its own commit.)

src/componentize.js
vados-cosmonic: Ah thanks, will take this change and leave the comment out!

vados-cosmonic:
Definitely can do -- I'll split this into more commits so it's a bit easier

vados-cosmonic:
Hey @tschneidereit so it turned into 8 commits :sweat_smile:  -- hopefully these changes are easier to follow.

vados-cosmonic:
Putting this in draft until it's rebased and ready to be looked at again to avoid taking attention from other PRs!

vados-cosmonic:
OK, tests are passing so I'm taking this out of draft -- it's back into one commit, which I think isn't too difficult to review now, given that some changes have already made it into `main`, the diff is much more tame.

This PR does two things:

- Expands the feature-set and granularity of "debug" options to `componentize-js`
- refactors the code to be a bit more readable.

src/componentize.js
```diff
@@ -89,89 +82,105 @@ export async function componentize(opts,
...
-    tmpdir(),
-    createHash('sha256')
-      .update(Math.random().toString())
-      .digest('hex')
-      .slice(0, 12)
-  );
-  await mkdir(tmpDir);
-  const sourceDir = maybeWindowsPath(join(tmpDir, 'sources'));
-  await mkdir(sourceDir);
+  // Prepare a working diretory for use during componentization
```
tschneidereit, line 92:
```suggestion
  // Prepare a working directory for use during componentization
```

src/componentize.js
```diff
@@ -89,89 +82,105 @@ export async function componentize(opts,
...
-        .map((ln, idx) => `${(idx + 1).toString().padStart(4, ' ')} | ${ln}`)
-        .join('\n')
-    );
-    console.log('--- JS Imports ---');
-    console.log(imports);
-    console.log('--- JS Exports ---');
-    console.log(exports);
+    // If a bindings output directory was specified, output generated bindings to files
+    if (debug?.bindingsDir) {
+      console.error(`outputting debug files to [${debug?.bindingsDir}]...\n`);
```
tschneidereit, line 140: "Outputting" sounds a bit jarring to me :)
```suggestion
      console.error(`Storing debug files in "${debug?.bindingsDir}"\n`);
```

src/componentize.js
```diff
@@ -89,89 +82,105 @@ export async function componentize(opts,
...
+      // Ensure the debug bindings dir exists, and is a directory
+      if (!(await stat(debug?.bindingsDir).then((s) => s.isDirectory()))) {
```
tschneidereit, line 142: Is there a strong reason to use the async version of `stat` here, instead of the sync one we could use without `await` and `then`?

src/componentize.js
```diff
@@ -89,89 +82,105 @@ export async function componentize(opts,
...
+        jsBindings
+          .split('\n')
+          .map((ln, idx) => `${(idx + 1).toString().padStart(4, ' ')} | ${ln}`)
+          .join('\n'),
+      );
+      console.error('--- JS Imports ---');
+      console.error(imports);
+      console.error('--- JS Exports ---');
+      console.error(exports);
+    }
```
tschneidereit, line 173: Do you think we should even still keep this output? I've never found it useful to look at this directly, and always prefer having files I can open. If you agree, maybe just remove this entirely?

src/componentize.js
```diff
@@ -215,11 +224,10 @@ export async function componentize(opts,
   env['IMPORT_CNT'] = imports.length;
 
   if (debugBindings) {
-    console.log('--- Wizer Env ---');
-    console.log(env);
+    console.error('--- Wizer Env ---');
+    console.error(env);
   }
```
tschneidereit, line 229: I think it'd be good to also write this to a file if `bindingsDir` was passed

src/componentize.js
```diff
@@ -301,122 +310,55 @@ export async function componentize(opts,
...
+  // If the wizer (or weval) process failed, parse the output and display to the user
+  if (postProcess.status !== 0) {
+    let wizerErr = parseWizerStderr(postProcess.stderr);
     let err = `Failed to initialize component:\n${wizerErr}`;
     if (debugBindings) {
-      err += `\n\nBinary and sources available for debugging at ${tmpDir}\n`;
+      err += `\n\nBinary and sources available for debugging at ${workDir}\n`;
     } else {
-      rmSync(tmpDir, { recursive: true });
+      await rm(workDir, { recursive: true });
```
tschneidereit, line 333: Are there benefits to switching to the async version here?

src/componentize.js
```diff
@@ -464,3 +418,172 @@ function defaultMinStackSize(freeMemoryBytes) {
...
+    );
+  };
+
+  let stderr = '';
+  const wasmModule = await WebAssembly.compile(bin);
+
+  const mockImports = {
+    // "wasi-logging2": {
+    //   log: eep("log"),
+    // },
```
tschneidereit, line 519:
```suggestion
```

tschneidereit (approved):
I think this all looks good, modulo a few comments and suggestions.

However, I didn't review the big functions at the end of `componentize.js`, because I _think_ they were just moved from elsewhere. If that's correct, please feel free to merge this. If not, it'd be great if you could still split out that code motion from the rest of the changes, so I can look at what actually changed instead of losing it in the code motion.

src/componentize.js
vados-cosmonic: Ah so this is actually something I wanted to keep for backwards compat -- I think the number of people currently expecting to hit this code path is small, but I didn't want to have someone's debug process stop working suddenly (or someone unfamiliar with the new debug options).

src/componentize.js
vados-cosmonic: Just for consistency here, similar to the other places -- to avoid using sync functions (that *should* complete quickly) with async.

PR #205, opened 2025-03-19
Don't expose absolute paths to source files

tschneidereit:
This changes componentization to make paths relative to the current working directory. E.g. when componentizing `[project dir]/src/index.js`, instead of seeing that full path, content will now see `src/index.js`, because we preopen `[project dir]` as `/` and can thus treat all paths as relative to that.

This only works for Wizer, since Weval doesn't have a way to map directories in this way.


tschneidereit:
(Note that this is based on #202, so only the last commit needs review here.)

guybedford (approved):

PR #206, opened 2025-03-20
feat: support min stack size for aot

vados-cosmonic:


src/componentize.js
```diff
@@ -236,6 +254,14 @@ export async function componentize(jsSource, witWorld, opts) {
       wevalBin = await getWeval();
     }
 
+    // Set the min stack size, if one was provided
+    if (opts.aotMinStackSizeBytes) {
+      if (!isNumeric(opts.aotMinStackSizeBytes)) {
+        throw new TypeError(`aotMinStackSizeBytes must be a numeric value, received [${opts.aotMinStackSizeBytes}] (type ${typeof opts.aotMinStackSizeBytes})`);
+      }
+      env.RUST_MIN_STACK = opts.aotMinStackSizeBytes;
+    }
```
guybedford, line 263: I find Weval often exceeds its own default limit such that it can be worth setting a higher value here by default. `Math.max(8 * 1024 * 1024, Math.floor(freemem() * 0.1))` seems to work, although I hven't done a lot of testing here.

vados-cosmonic: Hey so I was of the opinion here that the clients of componentize-js should set this value (to that or otherwise) -- so I [set it in jco](https://github.com/bytecodealliance/jco/pull/608/files#diff-82eb8bdf9791d41794262b9600e8c0a3520f0fc3a27c08c4b873d33fe3c9d780R36), what do you think? Open to moving that default here (and maybe only here?)

guybedford: Ah, yeah maybe worth upstreaming here especially now we have the CLI. It might also be worth considering migrating `jco componentize` to `componentize-js` as well if the ComponenizeJS CLI is now first class to avoid maintaining two separate API interfaces.

vados-cosmonic: OK great, will upstream (and rebase this on top of Till's changes that have just landed)

Hmnnn I'm a bit hesitant to migrate `jco componentize` to `componentize-js` just yet -- it doesn't feel like `componentize-js` is ready functionality/documentation/tests/examples/etc wise to point everyone to... Definitely worth talking through more though.

vados-cosmonic: Done!

PR #207, opened 2025-03-21
fix(ci): use a local test server for fetch test

vados-cosmonic:
This test seems to be flaky in CI mostly because it relies on a remote URL.

What I'd *like* to do here is use a server that *we* manage (similar to how `jco` does, moving over some utility functions from there) that will definitely return JS, but it's not clear how the exported `source` was intended to be used (is it OK for that to become a function that takes a URL?).

In the meantime, we can at least retry the test while it depends on a remote URL (and we don't have any testing frameworks like `vitest` set up).

### DIFF
--- a/test/builtins/fetch.js
+++ b/test/builtins/fetch.js
@@ -4,13 +4,11 @@ import { createServer } from 'node:http';
 import { strictEqual, ok } from 'node:assert';
 
 import { maybeWindowsPath } from '../../src/platform.js';
+import { getRandomPort } from '../util.js';
 
 const FETCH_URL = 'http://localhost';
 
 export const state = async () => {
-  const { getRandomPort } = await import(
-    maybeWindowsPath(fileURLToPath(new URL('../util.js', import.meta.url)))
-  );
   const port = await getRandomPort();
   return { port };
 };
--- a/test/builtins/fetch.js
+++ b/test/builtins/fetch.js
@@ -3,11 +3,13 @@ import { createServer } from 'node:http';
 
 import { strictEqual, ok } from 'node:assert';
 
+import { maybeWindowsPath } from '../../src/platform.js';
+
 const FETCH_URL = 'http://localhost';
 
 export const state = async () => {
   const { getRandomPort } = await import(
-    fileURLToPath(new URL('../util.js', import.meta.url))
+    maybeWindowsPath(fileURLToPath(new URL('../util.js', import.meta.url)))
   );
   const port = await getRandomPort();
   return { port };
--- a/test/builtins/fetch.js
+++ b/test/builtins/fetch.js
@@ -8,17 +8,11 @@ import { getRandomPort } from '../util.js';
 
 const FETCH_URL = 'http://localhost';
 
-export const state = async () => {
-  const port = await getRandomPort();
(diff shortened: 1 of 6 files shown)


tschneidereit:
ah, I didn't see this PR before filing #209. I think we should move away from that remote URL entirely instead of trying to paper over it. One reason is that it seems like the failures show up somewhat consistently for all tests of a run if they show up at all: I usually see all of `release`, `debug`, and `release-weval` failing together. That seems to indicate that retries wouldn't necessarily help.

vados-cosmonic:
Yeah, looks like we have the same intent there -- updating the code to use a temporary server we control is likely the best solution here, skipping the stopgap.

vados-cosmonic:
Hey @tschneidereit can you take a look at this? should be ready to go now

tschneidereit:
@vados-cosmonic thank you for this! It seems like there are still some Windows issues to address, but once those are dealt with, I'm eager to get this landed.

vados-cosmonic:
Ah apologies there -- windows paths were being wonky, but I didn't need to do that dynamic import anyway! Should be good now

test/builtins/fetch.js
```diff
@@ -1,20 +1,70 @@
...
+
+const FETCH_URL = 'http://localhost';
+
+export const state = async () => {
+  const port = await getRandomPort();
+  return { port };
+};
+
+export const source = (testState) => {
+  let port = testState?.port ? ':' + testState.port : '';
```
tschneidereit, line 17: This PR introduces a lot of machinery for handling `testState`, and I'm not entirely convinced it carries its own weight.

As far as I can tell, all of this could be removed in favor of a simple

`const port = await getRandomPort()` at this file's top-level, and then simply using that port in this and the `test` function. I would find that much easier to follow, and I can't really identify a reason why it wouldn't work just as well. If you feel that I'm wrong about this, at the very least there should be solid documentation around it all.

tschneidereit (commented):
I left a comment in-line on test state management. I'd like to discuss that more before landing this.

test/builtins/fetch.js
vados-cosmonic: So the reason I did this was that there are three places that interact:

- the test needs to know the port (to make the server)
- the compiled component needs to know the port (to access the server)

This state belongs to the test, but it's really per-test-run state, not really for the module. Today it's just the one port in one test, but having machinery to create necessary test-time state for a given test to ensure they can be easily parallelized and we don't have to worry about effects using source w/ a hardcoded same-port twice seemed like a good thing to do early.

This is something that came up while working on Jco's tests -- IIRC the simple example was testing the same module under AOT and non-AOT runs. If I did that, and loaded the module once, as far as I can tell the module would have the same hard-coded port right? Which would then force us to run serially.

It would certainly *work* to have the port be determined at load time of this module but I think it would cut out some flexibility.

Happy to take this out though, we can add this machinery later if need be.

vados-cosmonic: I've gone ahead and removed the state machinery for now -- I'm certainly happy to move this forward with out it!

tschneidereit (approved):
Thank you! <3

PR #217, opened 2025-04-08
Fix building on windows when using new syntax

karthik2804:
The new sytax uses cwd() to strip the prefix when building using original source location. We were not running it through maybeWindowsPath which led to file not being found during initialization. This PR fixes that and thus enables building on windows using the new syntax.


tschneidereit (approved):
Thank you!

PR #223, opened 2025-05-10
Allow customizing `wizer` path using `opts.wizerBin`

### LINKED ISSUE #222: Consider making `wizer` path configurable
Similar to `opts.wevalBin`, I propose adding `wizerBin`. Running downloaded binaries can cause issues with reproducibility, security, and compatibility - for example, the downloaded binary doesn't work on NixOS.

tomasol:
Fixes #222


tschneidereit (approved):
Thank you!

PR #226, opened 2025-06-26
feat(ci): add release automation

vados-cosmonic:
This commit adds support for automated releases to componentize-js, following the same setup that is in use by jco.

Coordinated with this PR, the following have been added to this repo:
- [x] `NPM_ACCESS_TOKEN` secret
- [x] `RELEASE_PAT` secret
- [ ] [`jco-release-bot`](https://github.com/jco-release-bot) as a repo member


cliff.toml
```diff
@@ -0,0 +1,87 @@
...
+  { message = '^perf', group = '<!-- 4 -->⚡ Performance' },
+  { message = '^refactor', group = '<!-- 2 -->🚜 Refactor' },
+  { message = '^style', group = '<!-- 5 -->🎨 Styling' },
+  { message = '^test', group = '<!-- 6 -->🧪 Testing' },
+  { message = '^release', skip = true },
+  { message = '^chore', group = '<!-- 7 -->⚙️ Miscellaneous Tasks' },
+]
+
+# filter out the commits that are not matched by commit parsers
+filter_commits = true
```
tschneidereit, line 66: Would it be possible to change this to instead of fully filtering things out collect them into a "Other Changes" section of the release notes? We're not consistently using conventional commits, and would have to talk about changing that

tschneidereit (approved):
I only looked at the core release parts in full detail to verify that it does what the PR says. Otherwise, I trust that the details of package name resolution and all that works correctly—or we find when testing this the first time :)

cliff.toml
vados-cosmonic: Ah certainly! I'll change this

PR #227, opened 2025-06-26
chore: add CHANGELOG.md

vados-cosmonic:
:sweat_smile:


tschneidereit (approved):
I guess we need to address the base case first, huh?

PR #232, opened 2025-06-27
fix(ci): remove packages prefix

vados-cosmonic:
NPM pack build worked, but build provenance path was wrong:

https://github.com/bytecodealliance/ComponentizeJS/actions/runs/15930310966/job/44937659195


tschneidereit (commented):
Are the changes to the cache-node-modules jobs related, or are they drive-by cleanups you did? Either way, could you explain them so I don't have to derive what they're about myself?

vados-cosmonic:
Ah sorry, those were wrong for the same reason as the build provenance path -- there is no `packages` folder, so I changed those as well while I was at it!

tschneidereit (approved):

PR #237, opened 2025-07-03
feat: initialize debugger on first call

andreiltd:
This enables using content debugger for scripts that don't target wasi/http. Depending on how StarlingMonkey has been configured, `maybe_init_debugger` can be a nop or perform actual debugger initialization. There is currently a triple line of defense to prevent accidental debugger initialization:

- StarlingMonkey build must be configured with ENABLE_JS_DEBUGGER=ON
- StarlingMonkey runtime arguments must include the `--enable-script-debugging` option
- The environment variable DEBUGGER_PORT must be defined

### DIFF
--- a/embedding/embedding.cpp
+++ b/embedding/embedding.cpp
@@ -1,4 +1,5 @@
 #include "embedding.h"
+#include "debugger.h"
 #include "builtins/web/performance.h"
 
 namespace builtins::web::console {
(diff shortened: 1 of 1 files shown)


andreiltd:
Hey! Good point. I think we need to choose where the debugger documentation should live and link to it from other projects. I would prefer to have it in StarlingMonkey mdbook if we decide to move forward with it: https://andreiltd.github.io/StarlingMonkey/

tschneidereit (approved):
LGTM! Agreed with Victor on documentation, but given that we don't really have documentation for JS debugging at all right now, I don't want to block landing this on documentation. So I'll land it now, and we should talk about docs in more detail separately.

PR #241, opened 2025-07-07
fix(ci): npm release artifact

vados-cosmonic:
While componentize-js packages can be now automatically released, the packages that come out of [the build](https://www.npmjs.com/package/@bytecodealliance/componentize-js/v/0.18.3-rc.3) are missing generated files and GHA attestations don't look like they're being recognized by NPM.

After "local" testing (via another repo), this commit addresses all the issues with the current release CI step. This commit:

- uses package-lock.json for runner cache
- removes a redundant release build
- enables attestation publishing for NPM
- uses more direct artifact name
- requires node version >= 22.x

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,7 +44,7 @@ jobs:
         id: cache-node-modules
         uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
         with:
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
Doing *one* more test on this branch to make sure things aren't different, but this is ready for review -- this is the commit:

https://github.com/bytecodealliance/ComponentizeJS/pull/241/commits/551407f62940595463e23304e883ce91206a22f2

The second commit (the one that starts with `debug(ci)` will be removed once I confirm the dry run is :heavy_check_mark: )

PR #247, opened 2025-07-10
fix: allow for use of both manual & fetchEvent based HTTP

vados-cosmonic:
This commit updates the componentize-js code to support both fetchEvent and manual based HTTP.

To make this possible we needed to re-introduce a few things:

- A feature that represents fetch-event
- Allow controlling removal of the the built-in fetch event
- Allow skipping checks for manual wasi:http/incoming-handler impls
- Add oxc-parser to detect manual impls and toggle fetch-event feature

The docs are also updated to note the changes to API/usage.

### DIFF
--- a/src/componentize.js
+++ b/src/componentize.js
@@ -121,12 +121,12 @@ export async function componentize(
   const engine = getEnginePath(opts);
 
   // Determine the default features that should be included
-  const features = DEFAULT_FEATURES.reduce((acc, f) => {
+  const features = new Set();
+  for (let f of DEFAULT_FEATURES) {
     if (!disableFeatures.includes(f)) {
-      acc.add(f);
+      features.add(f);
     }
-    return acc;
-  }, new Set());
+  }
 
   if (!jsSource && sourcePath) {
     jsSource = await readFile(sourcePath, 'utf8');


src/componentize.js
```diff
@@ -119,8 +120,36 @@ export async function componentize(
...
+    );
+    disableFeatures.push('fetch-event');
+  }
+
+  // Splice the bindigns for the given WIT world into the engine WASM
+  let { wasm, jsBindings, exports, imports } = splicer.spliceBindings(
     await readFile(engine),
+    features,
     witWorld,
     maybeWindowsPath(witPath),
```
vados-cosmonic, line 154:
```suggestion
    maybeWindowsPath(witPath),
    jsSource,
```

src/componentize.js
```diff
@@ -119,10 +120,39 @@ export async function componentize(
...
+    jsSource,
```
vados-cosmonic, line 155:
```suggestion
```

crates/splicer-component/src/lib.rs
```diff
@@ -17,12 +20,13 @@ impl Guest for SpidermonkeyEmbeddingSplicerComponent {
...
     fn splice_bindings(
         engine: Vec<u8>,
+        features: Vec<Features>,
         world_name: Option<String>,
         wit_path: Option<String>,
         wit_source: Option<String>,
         debug: bool,
     ) -> Result<SpliceResult, String> {
-        splice::splice_bindings(engine, wit_source, wit_path, world_name, debug)
+        splice::splice_bindings(engine, features, world_name, wit_path, wit_source, debug)
```
vados-cosmonic, line 29:
```suggestion
        splice::splice_bindings(engine, features, wit_source, wit_path, world_name, debug)
```

test/cases/fetch-event-server/test.js
```diff
@@ -0,0 +1,22 @@
...
+
+export async function test(instance) {
+  const server = new HTTPServer(instance.incomingHandler);
+  let port = await getRandomPort();
+  server.listen(port);
+
+  try {
+    const resp = await fetch(`http://localhost:${port}`);
+    const text = await resp.text();
+    strictEqual(text, 'Hello world!');
```
vados-cosmonic, line 18:
```suggestion
    strictEqual(text, 'Hello World!');
```

crates/spidermonkey-embedding-splicer/src/wit.rs
```diff
@@ -0,0 +1,22 @@
...
+
+impl std::str::FromStr for Features {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        match s {
+            "stdio" => Ok(Features::Stdio),
+            "clocks" => Ok(Features::Clocks),
+            "random" => Ok(Features::Random),
+            "http" => Ok(Features::Http),
```
andreiltd, line 18: Should this include `fetch-event` as well?

src/componentize.js
```diff
@@ -119,8 +120,36 @@ export async function componentize(
...
+  if (!jsSource && sourcePath) {
+    jsSource = await readFile(sourcePath, 'utf8');
+  }
+  const detectedExports = detectKnownSourceExportNames(sourceName, jsSource);
+
+  // If there is an export of incomingHandler, there is likely to be a
+  // manual implementation of wasi:http/incoming-handler, so we should
+  // disable fetch-event
+  if (
+    features.includes('http-server') &&
```
andreiltd, line 140: Did you mean just `http` here or `fetch-event`?

crates/spidermonkey-embedding-splicer/src/wit.rs
vados-cosmonic: Ah yes, it should -- thanks!

src/componentize.js
vados-cosmonic, line 140:
```suggestion
    features.includes('http') &&
```

crates/spidermonkey-embedding-splicer/src/bindgen.rs
```diff
@@ -104,6 +106,9 @@ struct JsBindgen<'a> {
     resource_directions: HashMap<TypeId, AbiVariant>,
 
     imported_resources: BTreeSet<TypeId>,
+
+    /// Features that were enabled at the time of generation
+    features: &'a Vec<Features>,
```
andreiltd, line 111: I think `Features` is a bit unfortunate name as this implies it is already a list, but that's generated right?

test/cases/fetch-event-server/test.js
```diff
@@ -0,0 +1,22 @@
...
+
+export async function test(instance) {
+  const server = new HTTPServer(instance.incomingHandler);
+  let port = await getRandomPort();
+  server.listen(port);
+
+  try {
+    const resp = await fetch(`http://localhost:${port}`);
+    const text = await resp.text();
+    strictEqual(text, 'Hello World!');
```
vados-cosmonic, line 18:
```suggestion
    strictEqual(text, 'Hello World');
```

crates/spidermonkey-embedding-splicer/src/bin/splicer.rs
```diff
@@ -69,18 +75,13 @@ enum Commands {
 ///    clocks,
 ///    random,
 ///    http,
+///    fetch-event,
 ///}
-fn map_features(features: &Vec<String>) -> Vec<Features> {
+fn map_features(features: &Vec<String>) -> Result<Vec<Features>> {
```
andreiltd, line 80: I think clippy would yap that this should be just `features: &[String]` 🤷

andreiltd (approved):
Thank you! I wouldn't mind additional tests, for example for pure components that you mention in the docs, but otherwise LGTM! 🚀

src/componentize.js
```diff
@@ -119,8 +120,36 @@ export async function componentize(
...
+    jsSource = await readFile(sourcePath, 'utf8');
+  }
+  const detectedExports = detectKnownSourceExportNames(sourceName, jsSource);
+
+  // If there is an export of incomingHandler, there is likely to be a
+  // manual implementation of wasi:http/incoming-handler, so we should
+  // disable fetch-event
+  if (
+    features.includes('http') &&
+    detectedExports.contains('incomingHandler')
```
t3hmrman, line 141:
```suggestion
    detectedExports.includes('incomingHandler')
```

vados-cosmonic, line 141:
```suggestion
    detectedExports.includes('incomingHandler')
```

vados-cosmonic:
Also @tschneidereit before you run into this.. I have already split out the PRs so they're a bit easier to review:

https://github.com/bytecodealliance/ComponentizeJS/pull/252
https://github.com/bytecodealliance/ComponentizeJS/pull/251
https://github.com/bytecodealliance/ComponentizeJS/pull/248

I'll do the required PR fixups regardless of which you merge in what order!

vados-cosmonic:
Hey @tschneidereit -- I've rebased this to `main` with all the changes that landed, this should be a very minimal changeset + test now.

tschneidereit (approved):
Let's get this landed. There are a couple of things I'd prefer to have different, but absolutely don't want to block getting this in on them:

It'd be nice if we could, on a best-effort basis, detect usage of `FetchEvent` when auto-disabling support for it. That'd be possible using oxc, but somewhat gnarly. Basically, it'd require walking the AST, finding all function calls to the identifier `addEventListener`, and checking if the first argument is the string literal `"fetch"`. Obviously there are ways to create both false positives and negatives, but those are unlikely enough for a warning not to be a problem, I think.

The more important thing is that I thought we had agreed to have the parsing happen in Rust instead of JS. That'd both better support using the splicer as a Rust dependency, and be faster.

Again, we should absolutely land this without addressing either of these!

src/componentize.js
```diff
@@ -116,24 +120,36 @@ export async function componentize(
...
-  if (!disableFeatures.includes('stdio')) {
-    features.push('stdio');
-  }
-  if (!disableFeatures.includes('random')) {
-    features.push('random');
-  }
-  if (!disableFeatures.includes('clocks')) {
-    features.push('clocks');
+  // Determine the default features that should be included
+  const features = DEFAULT_FEATURES.reduce((acc, f) => {
```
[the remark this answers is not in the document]
vados-cosmonic: Absolutely this change is worth putting in! I thought the same thing myself but was feeling functional that day. Will change

vados-cosmonic:
Made an issue so that we don't forget to come back to the fetch event detection to improve it!

https://github.com/bytecodealliance/ComponentizeJS/issues/253

PR #248, opened 2025-07-10
chore(ci): add clippy

vados-cosmonic:
Appeasing clippy is clearly not a goal, but it may be nice to have the code be a bit more consistent and enforcably so.


tschneidereit:
I'm in favor of adding a Clippy check, but let's make sure that we're judicious in which lints we enable, instead of manually overriding those we don't like in the code. We could copy the [configuration from Wasmtime](https://github.com/bytecodealliance/wasmtime/blob/main/Cargo.toml#L204), for example?

vados-cosmonic:
Sure, those seem like reasonable defaults

vados-cosmonic:
Wait, actually @tschneidereit thinking about it some more -- I'm not sure we should allow then re-add. `wasmtime` does a lot of really advanced stuff that clippy would get noisy about but I don't think componentize-js is in the same boat. IMO the default lints should mostly be in place here with a *few* exceptions.

How about we allow `too_many_arguments` instead

PR #252, opened 2025-07-11
refactor: splicer WIT and generated bindings

vados-cosmonic:
This commit refactors the WIT for the splicer, and moves the generated bindings to their own module file which is likely a bit easier to grok for most, and a good place to put extra functionality (e.g. custom impls for the generated types).

The code from this pr is split out from #247 for easy reviewing

### DIFF
--- a/crates/spidermonkey-embedding-splicer/src/bin/splicer.rs
+++ b/crates/spidermonkey-embedding-splicer/src/bin/splicer.rs
@@ -3,7 +3,7 @@ use clap::{Parser, Subcommand};
 use std::fs;
 use std::path::PathBuf;
 
-use spidermonkey_embedding_splicer::wit::Features;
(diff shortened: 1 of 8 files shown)


vados-cosmonic:
Hey @tschneidereit yeah so this one is the *most recent* PR -- the stylistic changes are from the introduction of clippy in https://github.com/bytecodealliance/ComponentizeJS/pull/248 -- I should have been more explicit about the PR order. 

I see how you got to this PR first, the order in [this comment](https://github.com/bytecodealliance/ComponentizeJS/pull/247#issuecomment-3061365146) is backwards

tschneidereit:
Leaving stacked PRs in draft certainly works, but just giving an explicit heads-up should also be enough. Just requiring reviewers to have a global view of all PRs when reviewing a single one isn't great.

PR #261, opened 2025-07-17
chore(docs): add comments to API type in README

### LINKED ISSUE #101: Document `witPath` option
The `witPath` option to the `componentize` function can either be a path to a single wit file or to a directory. If the path is a wit file then the wit package must not have any dependencies. Otherwise, if the path is a directory, the package's dependencies can be located in a `deps` directory.

This should be documented as it is not obvious that this is the behavior.

vados-cosmonic:
This commit fills out the API types in the README with the comments that are present in `types.d.ts`, to add more context for every option.

Resolves #101

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,18 +212,115 @@ object. Using the `fetchEvent` requires enabling the `http` feature.
 
 ```ts
 export function componentize(opts: {
-  sourcePath: string,
-  witPath: string,
-  worldName: string,
-  debug?: bool,
-  debugBuild?: bool,
-  engine?: string,
-  preview2Adapter?: string,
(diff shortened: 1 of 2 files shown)


vados-cosmonic:
Ah so there *were* changes -- the `debug` options that were recently added were not expressed in the types before

tschneidereit (approved):
As always, I'd prefer for pure code motion to be split into separate commits. I haven't read these changes in thorough detail, because as-is that'd be a whole bunch of work. But given that this is just changing documentation, that's not the end of the world.

It's somewhat unfortunate that this introduces a complete, manual duplicate of the documentation, which is highly likely to get out of date quickly. I'm not entirely sure what to do about that, though: all the places where I could see us adding a "update this other place if you update this" note would be pretty unlikely to be noticed when actually doing updates.

PR #262, opened 2025-07-17
refactor(tests): refactor expected globals test

vados-cosmonic:


test/builtins/globals.js
```diff
@@ -6,11 +6,117 @@ export const source = `
...
+    '[' + EXPECTED_GLOBALS.map((v) => `"${v}"`).join(', ') + ']\n',
```
tschneidereit, line 119: This looks like it isn't just a refactoring, but actually changes the expected output to prepend a `$` to each name?

vados-cosmonic: Ah nope, so that's the template syntax for JS! `${var}` -- a little hard to see with the quotes so close to each other!

So the reason I made this change is so that diffs in this test are easier to spot. Note that the output from the component hasn't changed, so we're still generating similar output to compare.

tschneidereit: Apologies for being dense on this—that obviously makes a ton of sense!

tschneidereit (approved):

PR #270, opened 2025-07-22
refactor: 'features' -> 'feature' in WIT & Rust component

### LINKED ISSUE #249: Rename Features enum in WIT
As noted on a recent PR, it would be nice if the `features` enum on the WIT side was named a bit more appropriately, because from code it *seems* like it represents a list. 

This should be a tiny change and a tiny PR -- perfect for new contributors. 

To start off, the WIT for the splicer should be changed, and generally following errors should suffice.

vados-cosmonic:
Closes #249


tschneidereit (approved):
Thank you!

PR #277, opened 2025-07-30
feat(tests): replace mocha with vitest

vados-cosmonic:
This PR moves us from `mocha` to `vitest` -- the diff is massive only because of `package-lock.json`, but checking the test output should be enough to confirm he functionality -- all tests are maintained (and now run in parallel) here and they all pass like before.

One thing that was weird is that `test:release` is evidently the exact same as `test`...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@bytecodealliance/componentize-js",
-  "version": "0.18.4-rc.1",
+  "version": "0.18.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@bytecodealliance/componentize-js",
-      "version": "0.18.4-rc.1",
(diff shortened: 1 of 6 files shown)


tschneidereit (approved):
The changes look good to me. What'd be nice for this kind of change is to add some reasoning to the description. I don't myself have any opinion on which is better. Is this mainly/purely down to performance, or something else?

vados-cosmonic:
Ah so performance (mostly via the parallelism, but `vitest` is certainly more tunable than `mocha`), robustness (ex. `vitest` has a bunch of [utilities like `waitUntil`](https://vitest.dev/api/vi.html#vi-waituntil), and can integrate with [GitHub actions](https://vitest.dev/guide/reporters.html#github-actions-reporter)). Even down to having better messages from failures (once we're using vitest's asserts).

PR #278, opened 2025-07-30
refactor(tests): split up test suites, update docs

vados-cosmonic:
When not in CI (on really slow machines), this PR makes the tests finish faster by separating one large suite into multiple (that can run in parallel under vitest).

Tests also run faster locally with these changes, as all the tests can run in parallel and even weval tests will pass normally on a modern machine.

### DIFF
--- a/test/vitest.ts
+++ b/test/vitest.ts
@@ -8,6 +8,14 @@ import { defineConfig } from 'vitest/config';
  */
 const TIMEOUT_MS = process.env.CI ? 240_000 : 120_000;
 
+/**
+ * On relatively modest machines in CI when weval is enabled,
+ * tests that use it (i.e. any test that does a build w/ enableAot) can
+ * suffer from CPU resource contention.
+ */
+const MAX_CONCURRENT_SUITES =
+  process.env.CI && process.env.WEVAL_TEST ? 1 : undefined;
+
 const REPORTERS = process.env.GITHUB_ACTIONS
   ? ['verbose', 'github-actions']
   : ['verbose'];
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
OK, so this PR is ready for review, but I'm a *little* dissatisfied with the consistency of CI -- running with more concurrency seems to have made tests more unstable. 

Going to re-run [the latest job](https://github.com/bytecodealliance/ComponentizeJS/actions/runs/16648152603/) a few more times to make sure it passes *most* of the time so that the average PR should pass without issues.

tschneidereit:
@vados-cosmonic same request I often have: can you update the description here to add some motivation for this? It's so very much easier to review PRs when I know which lens to look through. It's also very useful for looking at changes later on.

vados-cosmonic:
Hey @tschneidereit yeah it's been a while on this PR -- I believe I was trying to finish the switch to `vitest` in this PR, in particular a follow up to this PR: 

https://github.com/bytecodealliance/ComponentizeJS/pull/277

In particular this comment here:

https://github.com/bytecodealliance/ComponentizeJS/pull/277#issuecomment-3136080571

When not in CI (on really slow machines), this PR makes the tests finish faster by separating one large suite into multiple (that can run in parallel under `vitest`).

tschneidereit:
Yeah, my apologies for not getting to the review sooner. If anything though, this drives home why explanations for changes are so important: they're quite often hard to derive from the changes themselves after the fact.

Regarding the reasoning you mention, if I read [this comment](https://github.com/bytecodealliance/ComponentizeJS/pull/278#issuecomment-3142816538) correctly, that doesn't really apply anymore? If that's correct, do we even want this change still?

And if yes, please do update the PR description with an explanation of what the PR does, and why.

vados-cosmonic:
> Regarding the reasoning you mention, if I read https://github.com/bytecodealliance/ComponentizeJS/pull/278#issuecomment-3142816538 correctly, that doesn't really apply anymore? If that's correct, do we even want this change still?

Ah so the tests still run *faster*, they just don't run as fast as they *should have* in CI due to the weval failures when run concurrently, they still run much faster locally. Note that all the *non-weval* builds still finish faster.

Will update the PR description!

PR #285, opened 2025-08-29
Update starlingmonkey and bumps package version

karthik2804:
This PR pulls in the latest StarlingMonkey, which means we have to drop support for AoT using weval.


test/builtins/performance.js
```diff
@@ -21,8 +21,8 @@ export async function test(run) {
 
   ok(stderr.includes(' ms'));
   const time = Number(stderr.split(' ms')[0]);
-  // TODO: fix back to half a millisecond when Weval fix is added
-  if (time > 25) {
+
+  if (time > 0.5) {
```
karthik2804, line 25: Are we okay with bumping this to be 5ms? CI seems to fail on the linux runners

vados-cosmonic, line 25: Looks like this is still a bit tight -- I'm seeing as much as 3ms in some of the failures. This variability may have to do with tests running alongside each other, but IIRC performance tests now run on their own -- that might be worth checking into, and/or increasing the budget here to set a baseline

vados-cosmonic (commented):
Thanks for taking this on @karthik2804 !

tschneidereit (approved):
This all looks great to me, thank you! I left a few comments, questions, and suggestions, but nothing major.

I'll work with @andreiltd on getting a StarlingMonkey release done, so we can use that instead of a random commit.

README.md
```diff
@@ -347,12 +316,6 @@ npm run build
 
 Before being able to use `componetize-js` (ex. via `npm link`, from `jco`), you'll need to run:
```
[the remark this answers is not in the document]
vados-cosmonic: oh also there is a typo in here -- `componentize-js` :)

test/builtins/performance.js
tschneidereit: If we're currently seeing 3ms, I am 100% certain that we'd see test failures with 5ms very very soon.

I think there fundamentally isn't a signal to be had with this kind of test: if we want to have perf regression tests, we need to put in the work to actually do performance testing, statistical analysis of results with good p values and all that. Just having a threshold that we set to "this doesn't give us failing tests all that often" is not something that will ever give us something useful. It will over time cost us a lot of effort to handle failing tests that should pass, and will make us be less sensitive when handling tests that actually fail for good reasons.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bytecodealliance/componentize-js",
-  "version": "0.18.5",
+  "version": "0.19.0",
```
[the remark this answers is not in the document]
karthik2804: removed the commit

StarlingMonkey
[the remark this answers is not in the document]
andreiltd: I think this should be as simple as merging this PR: https://github.com/bytecodealliance/StarlingMonkey/pull/258 ?

test/builtins/performance.js
vados-cosmonic: Hmnnn yeah I'm not sure I agree here -- I don't think we should let perfect be the enemy of good here, time taken to run the tests going up an order of magnitude is certainly something I don't expect to happen just with a "slow runner", and I certainly haven't seen up until now. 

That said I'm happy with with whatever ya'll decide to go with here!

README.md
```diff
@@ -49,9 +49,6 @@ As more components are written in JavaScript, and there exist scenarios where mu
 
 Establishing this initial prototype as a singular flexible engine foundation that can be turned into a shared library is therefore the focus for this project.
 
-### Weval AOT Compilation
-
-
```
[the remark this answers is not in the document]
karthik2804: I am not really sure of the explanation. This is what I found for context https://bytecodealliance.zulipchat.com/#narrow/channel/459697-StarlingMonkey/topic/Updating.20Gecko.20version/near/527089464

test/builtins/performance.js
tschneidereit: How about this: we keep it at a lower threshold for now, but the moment we receive any false positives from it, we simply remove this entire check?

vados-cosmonic: Oh yeah that certainly seems reasonable, but looking back at the errors, I see the following:

> Error: Error: took more than half a millisecond - 158.72219276428223 ms
(one of the windows runs)

> Error: Error: took more than half a millisecond - 10.32882308959961 ms
(ubuntu run)

Yeah, it's probably better to remove it all together, there's too much variance on the runners

README.md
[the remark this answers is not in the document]
vados-cosmonic: Ah maybe @tschneidereit if you could help out with something worth adding here that would be great! 🙇 

That said, in my own limited understanding of it, a summary of Chris's explorations/findings might be enough?

karthik2804:
I believe I have addressed all the comments on the PR, I will keep an eye out for a new tag of StarlingMonkey and update the PR once that lands.

PR #288, opened 2025-09-04
fix(tests): race condition with server random port

### LINKED ISSUE #287: Intermittent test failures in CI
On the last merge to `main` we had an intermittent [test failure](https://github.com/bytecodealliance/ComponentizeJS/actions/runs/17460975965/attempts/1). The test passed on a manually triggered re-run.

@vados-cosmonic it seems like this is due to `getRandomPort` not waiting for the temporary server it [creates](https://github.com/bytecodealliance/ComponentizeJS/blob/main/test/util.js#L21) to be closed before reporting the port.

However, I think there's a fundamental race condition here: either we have the race seen in this test failure, or we have a potential race between this and another test, where that other test might steal the port between the temporary server shutting down and our test server starting up.

That second one is probably less likely, but I'm not sure. A way to avoid either would be to not create a temporary server at all: why not create the actual test server in the global script instead of the test function, and return its port once it got created?

vados-cosmonic:
This commit updates the tests that used `getRandomPort()` to perform their own random port getting and use the generated servers where appropriate, fixing a race condition on acquiring a temporary server port.

To be able to use `HTTPServer` instances with os-provided random ports from `@bytecodealliance/preview2-shim`, this PR updates updates the dev-dependency on preview2-shim to 0.17.4.

Resolves #287

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,17 @@
 {
   "name": "@bytecodealliance/componentize-js",
-  "version": "0.18.4",
+  "version": "0.18.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@bytecodealliance/componentize-js",
-      "version": "0.18.4",
+      "version": "0.18.5",
       "workspaces": [
         "."
       ],
       "dependencies": {
         "@bytecodealliance/jco": "^1.9.1",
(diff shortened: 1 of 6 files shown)


tschneidereit (approved):
Thank you!

PR #290, opened 2025-09-08
fix(ops): add catch-all for non-conventional commits

vados-cosmonic:


vados-cosmonic:
Hey @tschneidereit I fixed the config to make sure non-cc commits get in there, if we can get this in relatively quickly I'm happy to recreate that release (or you can click the button again :smile: really happy to see someone else using the automation!)

vados-cosmonic:
Ah so it's not that the commit message was problematic, it was the *absense* of the thing you actually wanted to release! I was wondering why you might be excited to release only a testing change... Then realized that can't be the reason noticed Karthik's comment wasn't showing up!

I tested locally, and with this, we should get *both* commit messages to properly show up next time

PR #298, opened 2025-10-07
Engodev/improve compile time

EngoDev:
The goal of this PR is to decrease the time it takes to compile a JS file into a WASM component.

## AI disclaimer
I used AI in the creation of this PR. Specifically the Cursor composer (Claude-4.5-sonnet and GPT-5) for:
* The serialization logic at: `crates/spidermonkey-embedding-splicer/src/bin/splicer.rs`
* The splicer testing at: `test/splicer.js`
* Modifying the `Makefile`
* Extracting logic from `src/componentize.js` to `src/splicer.js`

## TL;DR
The two optimizations I introduced are:
1. Upping the version of `wirm` (previously known as `orca_wasm`) to the latest version and enabling the `parallel` feature.
2. Giving the option to do bindings and WASI stubbing using the CLI version of `spidermonkey-embedding-splicer` to take advantage of the multi threading capabilities of `wirm`.

## Research

I found that most of the time for compiling the JS code into WASM is being used in `Module::parse` (A `wirm` function) and the hot path for it is this code right here:

image

`Module::parse` is called multiple times during compilation and every time its called the `code_sections` creation takes about 95% of the run time for `Module::parse` which was around 350ms for me. As the great guys over at `wirm` already did fantastic work by giving the option to parallelize this code the only thing left to do was to enable the feature `parallel` but because WASM doesn't support multi threading yet just enabling the `parallel` feature for `wirm` doesn't do anything.

So to get around that I added a new option `--splicer-bin` that will use the `spidermonkey-embedding-splicer` CLI instead of the compiled WASM module so it can take advantage of multi threading and that made it possible to go from ~4.5s compilation time to ~750ms compilation time. 

## Performance Numbers

All testing took place on a Apple M4 Pro with 48GB of RAM.

I used a simple JS file for my testing:
```javascript
export async function run(context) {
	return "Hello World!";
}
```

With a simple `WIT` world:
```wit
package test:entry;

world entry {
	export run: func(context: string) -> string;

	import wasi:random/random@0.2.7;
	import wasi:clocks/wall-clock@0.2.5;
}
```

Running `jco` on it took 4.5s on average.
image

Running `componentize-js` with the `wirm` up to date took 1.5s on average.
image

Running `componentize-js` with the `wirm` up to date and leveraging the CLI splicer for multithreading took 750ms on average.
image

## Question

This PR right now doesn't have anything in the README on `--splicer-bin`, I wanted to know first what is the approach you would want to take regarding the packaging of the CLI splicer, right now it's not published and in the npm pacakge there is only the WASM module. I thought of two ways we can tackle user consumption but if you have another approach I'm happy to implement :)
1. Add a section in the readme from beginning to end on how to use the `--splicer-bin` which will include compiling
2. Add cross compiling for the major platforms in CI to compile the CLI tool and package that into the npm package as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ wasmparser = { version = "0.239.0", features = ["features",
     "std",
     "validate",
 ] }
-
 wit-bindgen = { version = "0.41.0", features = [ "macros", "async", "realloc" ] }
 wit-bindgen-core = { version = "0.41.0", default-features = false }
 wit-component = { version = "0.227.1", features = ["dummy-module"] }
--- a/crates/spidermonkey-embedding-splicer/src/bin/splicer.rs
+++ b/crates/spidermonkey-embedding-splicer/src/bin/splicer.rs
@@ -208,7 +208,7 @@ fn serialize_imports(imports: &[(String, String, u32)]) -> String {
 /// Manually serialize CoreFn to JSON
 fn serialize_core_fn(core_fn: &CoreFn) -> String {
     let mut result = String::from("{");
-    
+
     // params
     result.push_str("\"params\": [");
     for (i, param) in core_fn.params.iter().enumerate() {
@@ -218,7 +218,7 @@ fn serialize_core_fn(core_fn: &CoreFn) -> String {
         result.push_str(&serialize_core_ty(param));
     }
     result.push_str("], ");
-    
+
     // ret
     result.push_str("\"ret\": ");
     if let Some(ref ret) = core_fn.ret {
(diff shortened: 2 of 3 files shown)


tschneidereit:
Thank you for looking into this!

If I understand your results correctly, most of the improvement comes from updating `wirm`, cutting the time by 66%, right? If so, would you be up for splitting that change off into its own PR? That seems uncontroversial, so we could land it quickly.

The change to use parallelization and hence moving to a native build seems much more involved: we'd have to publish native versions for at least three, ideally four or five platforms, same as we already do for Wizer. That's not the end of the world for sure, but if we get most of the improvement without such an invasive change, we should consider not doing it.

This is in part because we're planning on completely changing how we do bindings generation for StarlingMonkey, by adopting the new [wit-dylib](https://github.com/bytecodealliance/wasm-tools/pull/2304) tool. At that point, everything done to ComponentizeJS as it is right now will become moot, so I think we shouldn't stay away from involved changes.

EngoDev:
Yeah that makes sense, I'll open another PR with only the `wirm` update.

I do think it's worth considering the CLI change as it does add a significant performance boost for little to no effort and it's an opt in feature so users aren't bothered with it.

EngoDev:
@tschneidereit Opened a new PR https://github.com/bytecodealliance/ComponentizeJS/pull/300/files

PR #299, opened 2025-10-07
fix: separate folder component path use

vados-cosmonic:
Previously, when attempting to call `componentize()` on a with a `sourcePath` in a different directory, the call to Wizer would be generated with the incorrect *relative* path (path prefix stripping).

This commit ensure that when Wizer is called with path prefix stripping, the path to the mapped file is correct.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -153,7 +153,8 @@ jobs:
           # - latest reenable when https://github.com/nodejs/node/issues/57172 is fixed
         os:
           - ubuntu-latest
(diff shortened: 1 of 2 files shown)


vados-cosmonic:
Hey so I think there might be a problem with `windows-latest` -- ever since GitHub Actions switched from `windows-2022` to `windows-2025` this summer tests have hung in Jco (I had to manually revert to `windows-2022` rather than windows-latest`) -- what do you want to do @tschneidereit ? I don't think these tests will ever stop running.

PR #300, opened 2025-10-07
Updated orca_wasm into wirm latest version

EngoDev:
Split from https://github.com/bytecodealliance/ComponentizeJS/pull/298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Changelog
 
+## [0.19.2] - 2025-10-17
+
+### 🚀 Features
+
+* Update StarlingMonkey to v0.2.1 (#311) by @tschneidereit in #311
+
+
+### 🐛 Bug Fixes
+
+* *(ci)* test runs on windows runner (#308) by @vados-cosmonic in #308
+
+* separate folder component path use (#299) by @vados-cosmonic in #299
+
+* remove shell setting for wizer call (#303) by @vados-cosmonic in #303
+
+* *(ci)* temporarily downgrade windows runners from latest to 2022 (#302) by @vados-cosmonic in #302
+
+
+
+
 ## [0.19.1] - 2025-09-16
 
 ### 🐛 Bug Fixes
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@bytecodealliance/componentize-js",
-  "version": "0.19.0",
+  "version": "0.19.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@bytecodealliance/componentize-js",
-      "version": "0.19.0",
+      "version": "0.19.1",
       "workspaces": [
         "."
       ],
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bytecodealliance/componentize-js",
-  "version": "0.19.1",
+  "version": "0.19.2",
   "homepage": "https://github.com/bytecodealliance/componentize-js#readme",
   "description": "ESM -> WebAssembly Component creator, via a SpiderMonkey JS engine embedding",
   "type": "module",
--- a/src/cli.js
+++ b/src/cli.js
@@ -21,7 +21,7 @@ export async function componentizeCmd(jsSource, opts) {
 }
 
 program
-  .version('0.19.1')
+  .version('0.19.2')
   .description('Create a component from a JavaScript module')
   .usage('<js-source> --wit wit-world.wit -o <component-path>')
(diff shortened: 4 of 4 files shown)


Cargo.toml
```diff
@@ -14,10 +14,11 @@ too_many_arguments = 'allow'
 anyhow = { version = "1.0.95", default-features = false }
 heck = { version = "0.5", default-features = false }
 js-component-bindgen = { version = "1.11.0" }
-orca-wasm = { version = "0.9.2", default-features = false }
+wirm = { version = "2.1.0", features = ["parallel"] }
```
tschneidereit, line 17: This shouldn't use the `"parallel"` feature, and probably use `default-features = false` again (though I don't know what the default features are, and hence whether we could use them.)

crates/spidermonkey-embedding-splicer/src/bin/splicer.rs
```diff
@@ -139,8 +141,112 @@ fn main() -> Result<()> {
                     out_dir.join("initializer.js").display()
                 )
             })?;
+
```
tschneidereit, line 144: Can you say what the code added in this file is for? It doesn't seem like the resulting JSON files are read anywhere.

crates/spidermonkey-embedding-splicer/src/splice.rs
```diff
@@ -757,32 +757,50 @@ fn synthesize_import_functions(
...
-            if let Operator::I32Const { value: ref mut v } = instr.op {
-                // we specifically need the const "around" 3393
-                // which is the coreabi_sample_i32 table offset
-                if *v < 1000 || *v > 5000 {
-                    continue;
+        // Find the I32Const base index and compute the delta to new base
+        let mut table_instr_idx = 0usize;
+        let mut delta: i32 = 0;
+        {
+            let ops_ro = builder.body.instructions.get_ops();
```
tschneidereit, line 764: I think you should be able to revert the changes to this loop and instead change `instructions.iter_mut()` to `instructions.get_ops().iter_mut()` in (the original) line 762.

crates/spidermonkey-embedding-splicer/src/splice.rs
```diff
@@ -757,32 +757,50 @@ fn synthesize_import_functions(
...
+                    table_instr_idx = idx;
+                    break;
                 }
-                *v = import_fn_table_start_idx;
-                table_instr_idx = idx;
-                break;
             }
         }
-        builder.inject_at(
+        // Inject adjustments after the located instruction: add delta and add arg index
```
tschneidereit, line 778: Why doesn't the `inject_at` approach work here anymore? Maybe you just couldn't find it anymore because it moved into the `InjectAt` trait? If you change the `use wirm::opcode::Inject;` statement to `use wirm::opcode::{Inject, InjectAt};` I think the original code should work as before.

crates/spidermonkey-embedding-splicer/src/splice.rs
EngoDev: `builder.body.instructions.get_ops()` returns an immutable reference so you can't use `iter_mut()` on it

crates/spidermonkey-embedding-splicer/src/splice.rs
EngoDev: Yeah I totally missed the `InjectAt`, added it back but the other code is necessary to set the instruction as it can't be done in the loop for the reason in the previous comment

Cargo.toml
EngoDev: Correct, changed it back.

crates/spidermonkey-embedding-splicer/src/bin/splicer.rs
EngoDev: I moved it by accident from the previous PR, sorry about that. Removed

crates/spidermonkey-embedding-splicer/src/bin/splicer.rs
```diff
@@ -5,7 +5,9 @@ use std::str::FromStr;
 use anyhow::{Context, Result};
 use clap::{Parser, Subcommand};
 
-use spidermonkey_embedding_splicer::wit::exports::local::spidermonkey_embedding_splicer::splicer::Feature;
+use spidermonkey_embedding_splicer::wit::exports::local::spidermonkey_embedding_splicer::splicer::{
```
tschneidereit, line 8: This can also be reverted, right?

crates/spidermonkey-embedding-splicer/src/splice.rs
tschneidereit: Ah, right. There's a `get_ops_mut()` that should hopefully work, though

PR #304, opened 2025-10-09
feat: add OCI annotations

### LINKED ISSUE #301: Support OCI annotations custom metadata tool convention.
Similar to what [cargo-component does](https://github.com/bytecodealliance/cargo-component/pull/372), we want components produced by guest toolchains to adhere to this (not yet merged/accepted but in practiced used) tool convention: https://github.com/WebAssembly/tool-conventions/pull/248

Refer to wasm-metadata's [oci_annotations](https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-metadata/src/oci_annotations) module.

ricochet:
Fixes #301

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@bytecodealliance/componentize-js",
-  "version": "0.19.0",
+  "version": "0.19.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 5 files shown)


vados-cosmonic:
Closing this for now, since we do have a tracking issue, and the fix needs to actually be made upstream in jco. 

In the case that the new rust-based componentize-js makes it in before this is finished, the impl will also look completely different.

PR #306, opened 2025-10-13
fix: duplicated function names for same interfaces and export

SanderVanRiessen:
This PR is a fix for the following issue: https://github.com/bytecodealliance/ComponentizeJS/issues/307

### DIFF
--- a/crates/spidermonkey-embedding-splicer/src/bindgen.rs
+++ b/crates/spidermonkey-embedding-splicer/src/bindgen.rs
@@ -174,7 +174,7 @@ pub fn componentize_bindgen(
     let mut import_bindings = Vec::new();
     for (specifier, item) in bindgen.imports.iter() {
         // this import binding order matters
-        import_bindings.push(binding_name_import(
+        import_bindings.push(generate_binding_name_import(
             &item.resource.func_name(&item.name),
             &item.iface_name,
             specifier,
@@ -205,8 +205,11 @@ pub fn componentize_bindgen(
             let item = items.first().unwrap();
             if let Some(resource) = resource {
                 let export_name = resource.to_upper_camel_case();
-                let binding_name =
-                    binding_name_import(&export_name, &item.iface_name, &item.binding_name);
+                let binding_name = generate_binding_name_import(
+                    &export_name,
+                    &item.iface_name,
+                    &item.binding_name,
+                );
                 if item.iface {
                     specifier_list.push(format!("{export_name}: import_{binding_name}"));
                 } else {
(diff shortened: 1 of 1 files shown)


crates/spidermonkey-embedding-splicer/src/bindgen.rs
```diff
@@ -1294,6 +1299,21 @@ fn binding_name(func_name: &str, iface_name: &Option<String>) -> String {
...
+        .map(|c| if c.is_alphanumeric() { c } else { '_' })
+        .collect::<String>();
+
+    if import_name != "<<INVALID>>" {
+        format!("{valid_import}${func_name}")
+    } else if let Some(iface_name) = iface_name {
+        format!("{iface_name}${func_name}")
+    } else {
+        func_name.to_string()
+    }
```
vados-cosmonic: Since `valid_import` doesn't get used unless we're in this case, maybe we can move the calculation in here?

Also, some comments on the expected cases would be great.

vados-cosmonic (commented):
Hey @SanderVanRiessen thanks for this addition -- it looks good to me, but I'd love some tests here to back the new functionality, and regression-proof the old functionality if you don't mind!

vados-cosmonic:
BTW, please ignore the windows CI failures -- we're unfortunately having problems with the windows jobs in CI right now.

vados-cosmonic:
CI failures from the windows runs should be good to go now!

crates/spidermonkey-embedding-splicer/src/bindgen.rs
vados-cosmonic: OK looks like this got changed!

crates/spidermonkey-embedding-splicer/src/bindgen.rs
```diff
@@ -1294,6 +1299,20 @@ fn binding_name(func_name: &str, iface_name: &Option<String>) -> String {
     }
 }
 
+fn binding_name_import(func_name: &str, iface_name: &Option<String>, import_name: &str) -> String {
```
vados-cosmonic: What do you think about adding a comment here? Something like the following:

```
/// Determine the binding name of a given import
///
/// # Arguments
/// * `func_name` - function name (e.g. `greet`)
/// * `iface_name` - an interface name, if present (e.g. `
/// * `import_name` - qualified import specifier
fn binding_name_import(func_name: &str, iface_name: &Option<String>, import_name: &str) -> String {
```

SanderVanRiessen: Hi, I fixed this in the latest commit.

vados-cosmonic: Thanks!

crates/spidermonkey-embedding-splicer/src/bindgen.rs
```diff
@@ -654,11 +655,11 @@ impl JsBindgen<'_> {
         let fn_name = func.item_name();
         let fn_camel_name = fn_name.to_lower_camel_case();
 
-        use binding_name as binding_name_fn;
+        use binding_name_import as binding_name_fn;
```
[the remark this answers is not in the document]
vados-cosmonic: Yeah, for those who might be watching -- trying to use the rename on the exports breaks *a lot* of things and results in a bug at a lower level (StarlingMonkey/the embedding?) -- much more investigation is needed. I spent a bunch of time on this yesterday and couldn't root cause a very weird bug -- even though the component would successfully *get into* the host-provided import, and run through it completely, it would error basically at the end of the function call (suggesting an engine level problem, somehow).

PR #308, opened 2025-10-17
fix(ci): test runs on windows runner

vados-cosmonic:
This commit fixes an somewhat odd problem with vitest on windows runners in that it cannot process properly `await import(...)` statements.

The changes in this commit represent the minimal changes to make the tests run properly instead of hanging, though there are other improvements to CI forthcoming.

### DIFF
--- a/test/bindings.js
+++ b/test/bindings.js
@@ -22,16 +22,18 @@ suite('Bindings', async () => {
         'utf8',
       );
 
-      const test = await import(`./cases/${name}/test.js`);
+      // NOTE: import separated from await due to issues on windows (see note in util.js)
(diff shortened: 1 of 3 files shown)


vados-cosmonic:
Yeah so what's really weird is that theoretically this code shouldn't even be converted by `tsc` per-say since they're all JS files -- the bug here might be in the transformation at the `vite` level -- so it might trace to `esbuild` at the end of the day. 

One thing I didn't try was slotting in something like `swc` instead of `esbuild`, but it also wasn't easy to find how to even configure `vite` through `vitest` -- I guess a `vite` config file in the right place might do it, maybe.

PR #319, opened 2026-02-25
fix: use-after-free in cabi_realloc free_list on repeated export calls

### LINKED ISSUE #224: Out of bounds memory access when using fetch
Originally opened in the jco repository (https://github.com/bytecodealliance/jco/issues/668). Also posting here for visibility. Feel free to close one of them.

---

Here is a minimal reproduction:

##  Create the following files

wit/component.wit:
```wit
package pack:name;

world component {
  import wasi:http/outgoing-handler@0.2.0;
  export wasi:cli/run@0.2.0;
}
```

src/main.js:
```js
const run = {
    async run() {
        try {
            var n = 0;
            let requests = [];
            while (n < 1000) {
                requests.push(fetch("https://jsonplaceholder.typicode.com/todos/1"));
                n += 1;
            }
            for (const request of requests) {
                const response = await request;
                const data = await response.json();
                console.log("HTTP Response:", data);
            }
        }
        catch (error) {
            console.error("HTTP Request Error:", error);
        }
    }
};

export { run };

```

## Run the following commands:

* `jco componentize -w wit -o dist/main.wasm src/main.js`
* `wasmtime run -S http=y dist/main.wasm `

## Output:
```
HTTP Response: { userId: 1, id: 1, title: "delectus aut autem", completed: false }
HTTP Response: { userId: 1, id: 1, title: "delectus aut autem", completed: false }
HTTP Response: { userId: 1, id: 1, title: "delectus aut autem", completed: false }
HTTP Response: { userId: 1, id: 1, title: "delectus aut autem", completed: false }
HTTP Response: { userId: 1, id: 1, title: "delectus aut autem", completed: false }
HTTP Response: { userId: 1, id: 1, title: "delectus aut autem", completed: false }
HTTP Response: { userId: 1, id: 1, title: "delectus aut autem", completed: false }
HTTP Response: { userId: 1, id: 1, title: "delectus aut autem", completed: false }
HTTP Response: { userId: 1, id: 1, title: "delectus aut autem", completed: false }
Error: failed to run main module `dist/main.wasm`

Caused by:
    0: failed to invoke `run` function
    1: error while executing at wasm backtrace:
           0: 0x7fcce3 - <unknown>!<wasm function 13434>
           1: 0x242dac - <unknown>!<wasm function 5248>
           2: 0x80c526 - <unknown>!post_wasi:cli/run@0.2.0#run
    2: memory fault at wasm address 0x9ad94527 in linear memory of size 0xb10000
    3: wasm trap: out of bounds memory access

Exception: wasmtime exited with 134
```

chaynabors:
Fixes #224

`cabi_realloc` tracked all allocations in `Runtime.free_list`, which `post_call` freed after each export invocation. When the host calls `cabi_realloc` during an import to write a return value into guest memory, those allocations may still be referenced by live JS objects across repeated export calls. `post_call` would free them, causing use-after-free on the next invocation.

This PR removes indiscriminate tracking from `cabi_realloc`. Only the `retptr` allocated explicitly in `call()` is tracked and freed by `post_call`. `retptr` is a temporary buffer that JS never references directly, which is why it needs to be managed explicitly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -184,19 +184,13 @@ jobs:
 
       - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
         with:
-          node-version: ">=22"
-
-      - name: Add npmrc
-        run: |
-          echo '//registry.npmjs.org/:_authToken=${NPM_TOKEN}' > .npmrc
+          node-version: ">=24"
 
       - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
         with:
           path: artifacts
 
       - name: Publish componentize-js to NPM
-        env:
-          NPM_TOKEN: ${{ secrets.NPM_ACCESS_TOKEN }}
         shell: bash
         run: |
           export PACKAGE_FILE_PATH=${{ github.workspace }}/artifacts/artifact/${{ needs.meta.outputs.artifact-name }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## [0.19.4-rc.1] - 2026-03-10
+
+### 🚀 Features
+
+* *(ci)* enable trusted publishing (#328) by @vados-cosmonic in #328
+
+
+
+
+## [0.19.4-rc.0] - 2026-03-10
+
+### 🚀 Features
+
+* support easy use of custom wizer bin (#323) by @vados-cosmonic in #323
+
+
+### ⚙️ Miscellaneous Tasks
+
+* update codeowners (#326) by @vados-cosmonic in #326
+
+
+
+
 ## [0.19.3] - 2025-10-27
 
 ### 🐛 Bug Fixes
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @tschneidereit @vados-cosmonic
+* @tschneidereit @vados-cosmonic @andreiltd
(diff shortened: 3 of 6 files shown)


chaynabors:
Accidentally closed #318 while doing some spring cleaning. Sorry for duplicate.

andreiltd:
Hey @chaynabors , sorry for the delay. I think this is pragmatic approach to fix double-free issue. I wonder how difficult would be to add the test that replicates https://github.com/bytecodealliance/ComponentizeJS/issues/224 as part of the fix? wdyt?

The next step should be to eliminate leaks by implementing proper deallocation in the splicer but that we can track in a separate issue.

chaynabors:
I think I had excluded the test because it was only reproducible under memory pressure, and because it was non deterministic, but that could have been a separate repo entirely. I'll take another look today and give an opinion.

I'll need to investigate again, but will also create an issue for the leak.

chaynabors:
Summary after re-reviewing:

The underlying issue is that import return buffers allocated via cabi_realloc in the splicer have no deallocation mechanism. This is now being tracked in #330.

The old free_list bulk free in post_call was the only thing reclaiming them, but it also freed buffers still in use. This PR narrows post_call to only free retptrs which should be the correct behavior long term.

I tried all morning to reproduce the issue in the test scaffolding but couldn't. When running my test with wasmtime directly I was able to reproduce on the first try. There's something about the jco execution environment that prevents the allocator from reusing freed regions the way wasmtime does. I don't know what that is.

When running the code in the issue with wasmtime, unpatched fails and patched succeeds. Patched also succeeds a comprehensive test suite in our own code, whereas unpatched traps almost immediately with an out of bounds memory access.

Happy to merge as written. I'm tuckered out.

chaynabors:
@tschneidereit Is it impolite to ask you to sanity check me here? You merged the original bookkeeping.

PR #333, opened 2026-04-14
fix: improve weval and wizer error reporting

andreiltd:
It looks like CI on macos sporadically fails under the load. This patch should improve error reporting in that case.

### DIFF
--- a/src/componentize.js
+++ b/src/componentize.js
@@ -281,6 +281,7 @@ export async function componentize(
   }
 
   let postProcess;
+  let postProcessName;
   if (opts.enableAot) {
     // Determine the weval bin path, possibly using a pre-downloaded version
     let wevalBin;
@@ -302,6 +303,7 @@ export async function componentize(
       env.RUST_MIN_STACK = defaultMinStackSize();
     }
 
+    postProcessName = `weval (${wevalBin})`;
     postProcess = spawnSync(
       wevalBin,
       [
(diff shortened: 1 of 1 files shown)


src/componentize.js
```diff
@@ -345,9 +348,15 @@ export async function componentize(
   }
 
   // If the wizer (or weval) process failed, parse the output and display to the user
-  if (postProcess.status !== 0) {
-    let wizerErr = parseWizerStderr(postProcess.stderr);
-    let err = `Failed to initialize component:\n${wizerErr}`;
+  if (postProcess.status !== 0 || postProcess.signal) {
+    let procErr = parseWizerStderr(postProcess.stderr);
+    if (postProcess.signal) {
+      procErr += `\nProcess was killed by signal: ${postProcess.signal}`;
```
vados-cosmonic, line 354: can we be specific about which one it was here? either wizer or weval?

vados-cosmonic (changes requested):
LGTM 🚀

src/componentize.js
andreiltd: This is done few lines below:

```js
let err = `Failed to initialize component (${postProcessName}):\n${procErr}`;
```

vados-cosmonic: Ah thank you -- I missed this

vados-cosmonic (approved):
LGTM 🚀 

(looks like GitHub doesn't let you approve and leave changes any more? Or did I misclick -- meant to approve previously)

PR #337, opened 2026-05-15
chore(deps): update js-component-bindgen

vados-cosmonic:


tschneidereit:
@vados-cosmonic there seems to be a lot more going on here than just an update of one dependency. Can you update the description to include which changes you made, so I know what to look for?

vados-cosmonic:
Hey @tschneidereit so the update of one dep triggered the other updates. Because a newer version of `js-component-bindgen` uses newer versions of ecosystem crates (and we can't use arbitrarily different versions) it prompts an update of the crates. Another thing that `componentize-js` uses is `wirm`, which *also* has a copy of the ecosystem crates -- so you can see where the updates come from.

That said, this isn't working yet -- there seems to be some generation problem I'm looking into, so it's not quite ready for review just yet -- but the intention here is definitely to update the single dep, it's just more involved than one would otherwise think. Two people from the community have already tried (and had a hard time).

vados-cosmonic:
Hey @tschneidereit OK this is ready for review now -- tests are green! 

Given the current code, the only change that will be made is `js-component-bindgen` which is @ `1.19.0-rc.7` will change to `1.19.0` as soon as [the release goes out](https://github.com/bytecodealliance/jco/pull/1502), no more changes to the code in `componentize-js` should be necessary.

crates/spidermonkey-embedding-splicer/src/splice.rs
```diff
@@ -471,7 +481,7 @@ fn synthesize_import_functions(
         .unwrap();
 
     // Sets the return value on args from the stack
-    let args_ret_i32: Vec<Operator> = vec![
+    let args_ret_i32: Vec<wirm::wasmparser::Operator> = vec![
```
tschneidereit: it looks like here and in many uses of `MemArg` and `Operator` below the added path qualifications aren't required, because `wirm::wasmparser::{Operator, MemArg}` are `use`d above.

crates/spidermonkey-embedding-splicer/src/stub_wasi.rs
```diff
@@ -2,8 +2,7 @@ use std::collections::HashSet;
 use std::path::PathBuf;
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use anyhow::{Result, bail};
-use wasmparser::{MemArg, TypeRef};
```
tschneidereit, line 6: Instead of removing this `use` statement, you can updated it to `wirm::` and drop the qualifications below.

rustfmt.toml
```diff
@@ -0,0 +1 @@
+edition = "2024"
```
tschneidereit: Shouldn't this be in the `Cargo.toml` file, in which case we wouldn't need a `rustfmt.toml`?

tschneidereit (approved):
Seems good, with two changes: the inline comments addressed and the commit message being a bit more descriptive, and the update to 2024 edition documented (with the introduction of `rustfmt.toml` either reverted or explained.)

crates/spidermonkey-embedding-splicer/src/splice.rs
vados-cosmonic: Ah good point -- will update these

rustfmt.toml
vados-cosmonic: `rustfmt` settings don't go in the `Cargo.toml`, they go in `rustfmt.toml` -- I added this because I get warnings, as `rust-analyzer` in my editor runs `rustfmt` as a [standalone program (rather than `cargo fmt`)](https://github.com/rust-lang/rustfmt/issues/4074).

That said, I'm happy to remove this

crates/spidermonkey-embedding-splicer/src/stub_wasi.rs
vados-cosmonic: Yup, good point, will do

rustfmt.toml
tschneidereit: understood re `rustfmt` settings. I would expect this particular setting to default to whatever is set in `Cargo.toml`, but maybe that's wrong? If it is, never mind

tschneidereit: Ah, and now I see that `Cargo.toml` already sets the same edition. So, never mind indeed :)

vados-cosmonic: Ah sorry to be clear, tooling *should* pick up on the edition in `Cargo.toml` (and it does when `cargo fmt` is used) -- this is really a specific problem for when anything runs `rustfmt` rather than `cargo fmt`, unfortunately :(

vados-cosmonic:
Sadly, the CI timeout needed to be bumped @tschneidereit @andreiltd . Looks like Windows + Weval is the new lagging test, so note for the future I guess. 

[EDIT] Also, *don't* merge this just yet! I will release `js-component-bindgen` 1.19.0 soon :)

vados-cosmonic:
OK, this is finally ready to go -- found some more places where I could update imports and remove qualified uses that I missed. `js-component-bindgen` is updated to 1.19.0, so once tests are :green_circle: again I'll merge.

PR #348, opened 2026-08-13
fix: use STARLINGMONKEY_SRC variable in include_directories

cuban-goats:
Line 18 hardcoded `"StarlingMonkey"` as the include path while the rest of the file consistently uses the `STARLINGMONKEY_SRC` variable, which also respects the `STARLINGMONKEY_SRC` environment variable override. This aligns `include_directories` with the existing pattern.


andreiltd (approved):
Thank you!

